PR #15, opened 2016-10-14
Add steam fallback to build.

mattfeldman:
Maintains preference for the current path, adds fallback to a common steam path. You could append more like this line in the future to support other paths if needed.


Pathoschild:
Looks good. Thanks for the contribution. :)

PR #37, opened 2016-12-10
Improve Controller Support and Make Scroll Buttons Functional

ShadowKitty42:


ShadowKitty42:
Wait.. that code I just removed, is required for another feature I made, which is making the scroll buttons in the lookup window clickable. I'll have to combine them with this.

Pathoschild:
Merged into `develop` for the upcoming 1.9 release. Thanks!

PR #68, opened 2017-05-09
enable built in debug menu

miketweaver:
enable the built in debug menu while the Mod's debug menu is up.


Pathoschild:
Merged into develop for the upcoming release. Thanks!

PR #77, opened 2017-07-07
ChestsAnywhere: add option to stack items

amireh:
This patch extends the ChestsAnywhere mod to allow the player to push a
button and have all the items in her inventory that already have stacks
in her chests be "stacked" with them.

I've noticed that quite a lot of my time playing the game is spent on
putting each item in the respective chest after looting, which is
irritating to the point that spoils the enjoyment of the game (to me, at
least.)

With this option, it only takes a single click (or keypress) to combine
items found in the inventory with existing stacks that have been placed
in the chest before-hand.

Future Improvements
-------------------

I'd like to be able to assign "specialties" to each chest, like
Vegetables or Fish, and have the stacking functionality utilize this
piece of data such that it can stack items even if they are not present
in the chest already. This would be useful for new seasons where you're
foraging and fishing new items, in which case it will reduce the manual
labor.

Problems
--------

- I couldn't find a proper place for the button so right now it's to the
  side of the existing Organize button
- For some reason, the hover text is not being honored in the chest
  menu; it would be nice to have a tooltip for the button since it may
  not be very intuitive. (Granted, this seems to be the case for the
  existing Sort/Organize button, so maybe we can fix both...)

### DIFF
--- a/ChestsAnywhere/ChestInventoryStacker.cs
+++ b/ChestsAnywhere/ChestInventoryStacker.cs
@@ -1,0 +1,85 @@
+﻿using System;
+using System.Linq;
+using System.Collections.Generic;
+using Pathoschild.Stardew.ChestsAnywhere.Framework;
+using StardewValley;
+
+namespace Pathoschild.Stardew.ChestsAnywhere
+{
+    internal class ChestInventoryStacker
+    {
+        public static void StackItemsInInventory(List<ManagedChest> chests)
+        {
+            // Combine an item stack with another of the same type.
+            //
+            // Returns whether the item stack has been completely combined.
+            Func<Item, Func<Item, bool>> combineItems = item => itemInChest =>
+            {
+                int space = item.getStack();
+                int availableSpace = Math.Min(space, itemInChest.getRemainingStackSpace());
+
+                if (availableSpace > 0)
+                {
+                    itemInChest.addToStack(availableSpace);
+                    item.addToStack(-1 * availableSpace);
+                }
+
+                return Math.Max(0, space - availableSpace) == 0;
+            };
+
+            // Index all available items in all chests for O(1) access later.
+            var itemsInChests = chests
+                .Aggregate(new List<Item> { }, (list, managedChest) =>
+                {
+                    return list.Concat(managedChest.Chest.items).ToList();
+                })
+                .Aggregate(new Dictionary<string, List<Item>> { }, (map, item) =>
+                {
+                    if (!map.ContainsKey(item.Name))
+                    {
+                        map.Add(item.Name, new List<Item> { });
+                    }
+
+                    map[item.Name].Add(item);
+
+                    return map;
+                })
+            ;
+
+            // Player inventory items; the check is necessary because empty items are included
+            var items = Game1.player.items.Where(x => x != null);
+
+            // Items that have an existing stack in some of the chests
+            var eligibleItems = items.Where(item => (
+                itemsInChests.ContainsKey(item.Name) &&
+                itemsInChests[item.Name].Any(itemInChest => itemInChest.canStackWith(item))
+            ));
+
+            var moves = eligibleItems.Select(item =>
+            {
+                return new
+                {
+                    item,
+                    // keep combining the stack until it is drained, if possible:
+                    drained = (
+                        itemsInChests[item.Name]
+                            .Where(itemInChest => itemInChest.canStackWith(item))
+                            .Any(combineItems(item))
+                    )
+                };
+            });
+
+            // Remove the items that were completely stacked from the player's inventory.
+            moves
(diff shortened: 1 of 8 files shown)


amireh:
Thanks for the great mod! The game would've been unplayable without it. Let me know if there's anything incorrect you can spot, I'd be happy to iterate.

Also, if you have any idea how to fix any of the problems I mentioned, please tell me.

(PS: I've only tested this on Linux, do we need to test on other platforms?)

amireh:
Here's a short video showing the extension in action: http://amireh.net/files/SV__ChestsAnywhere--Stacking.mp4

Pathoschild:
Hi @amireh. Thanks a lot for the contribution!

I like the [Unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy) for mods — each mod should do one thing and do it well, and Chests Anywhere is designed for compatibility with other mods. Chests Anywhere + [Chest Pooling](http://community.playstarbound.com/threads/smapi-0-40-chest-pooling-gateopener-toolcharging.111988/) + [Categorize Chests](http://www.nexusmods.com/stardewvalley/mods/1300) cover the features you mentioned.

I really like the button you added, and It'd be awesome if you released it as a separate stack-to-chests mod. (Feel free to copy code from Chests Anywhere as needed per the `LICENSE` file.) If you do, I can add a link to it as a recommended mod to use with Chests Anywhere. :)

amireh:
Thanks for the input! I totally agree about the separation of concerns. I actually did start it out as a standalone mod that depends on ChestsAnywhere (with your codebase as a starting point) but soon reverted to this approach for a few reasons:

1. having a coherent UX for the user where they don't have to go through a bunch of interfaces to configure each mod; things like the placement of the stacking button which has to accommodate for the Sort button provided by ChestsAnywhere. And then the functionality I wanted to add to the "Edit" page of the chest like specialty.
2. maintenance overhead; if the new mod has a hard dependency on yours, it's going to be painful for me to maintain since they'd be tightly coupled. I'm not sure I can afford the time for that...

Is it even possible to hook into the ManagedChest APIs provided by your mod in mine?

Finally, regarding the combination you mentioned, I still would have to go through each chest and push "stack" and I'm far too lazy to do that :grinning: 

Thanks again!

Pathoschild:
For the button position, you could do something like this:
```cs
// Default to the same position as Chests Anywhere's sort button.
// If Chests Anywhere is loaded, move button to the right.
Rectangle buttonBounds = new Rectangle(okButton.bounds.X, (int)(okButton.bounds.Y - sprite.Height * zoom - 5 * zoom), (int)(sprite.Width * zoom), (int)(sprite.Height * zoom));
if(helper.ModRegistry.IsLoaded("Pathoschild.ChestsAnywhere"))
   buttonBounds.X += (sprite.Width * zoom) + (5 * zoom);
```

You don't need to depend on Chests Anywhere, since you don't need most of its logic. You could use this simplified `ChestFactory` in your code to find chests for item stacking:
```cs
/// <summary>Encapsulates logic for finding chests.</summary>
internal class ChestFactory
{
    /*********
    ** Public methods
    *********/
    /// <summary>Get all player chests that items can be stacked into.</summary>
    public IEnumerable<Chest> GetChestsForStacking()
    {
        return (
            from chest in this.GetChests()
            where
                (chest.Name == null || !chest.Name.ToLower().Contains("|ignore|")) // ignore chests marked as 'hide this chest' in Chests Anywhere
                && chest.items.Any() // ignore empty chests
            select chest
        );
    }


    /*********
    ** Private methods
    *********/
    /// <summary>Get all player chests.</summary>
    Private IEnumerable<Chest> GetAllChests()
    {
        foreach (GameLocation location in Game1.locations)
        {
            // chests in location
            foreach (Chest chest in location.Objects.Values.OfType<Chest>())
                yield return chest;

            // chests in constructed buildings
            if (location is BuildableGameLocation buildableLocation)
            {
                foreach (Building building in buildableLocation.buildings)
                {
                    if (building.indoors == null)
                        continue;
                    foreach (Chest chest in building.indoors.Objects.Values.OfType<Chest>())
                        yield return chest;
                }
            }

            // farmhouse fridge
            if (location is FarmHouse house && Game1.player.HouseUpgradeLevel > 0)
            {
                Chest fridge = house.fridge;
                if (fridge != null)
                    yield return fridge;
            }
        }
    }
}
```

That way your mod works fine on its own, and plays well with Chests Anywhere if the player has that installed too.

amireh:
`helper.ModRegistry.IsLoaded` is super handy! Thank you for the informative response.

I'll take that approach then. In case I want to add the ability to specialize chests, we can just add a new menu like you do for editing a chest. I think that'll be fine. :grinning:

Thanks again for your work, it's top-quality stuff.

PR #83, opened 2017-07-16
Russian translate for LookupAnything

SteaNN:
Hello, this is translation for yours greatest mod to russian language.

All remarks can be forwarded to me.


Pathoschild:
Added to the next release. Thanks!

PR #102, opened 2017-08-26
Add Automation for Slime Incubator and Coop Incubator

danvolchek:
Adds a slime incubator machine and coop incubator machine. Sets the state of the new machines to never be done and GetOutput to always return null because when the machines are finished, they spawn a slime/chicken and then go back to being empty - they don't produce an output object.

AnimalHouse contains an incubator method, which seems to also incubate coop eggs. I tried making a coop incubator machine using that, but it was not functional - it looks unused and old.


Pathoschild:
Merged into `develop` for the upcoming Automate 1.7 release. Thanks!

Issue #62, opened 2017-04-20
[Fast Animations] Speed Up Leveled Water Can Animations


T-Miner:
I would love to see the multiple stages of the watering can animation be sped up. The watering can should just use the widest area quickly as possible.

Denifia:
I like the idea but always using the widest area possible would use a large amount of water each time. If you only need to water one square, it shouldn't use water for 9 squares.

Maybe just start with speeding up the animation and seeing if that is enough?

ghost:
@mralbobo's [tool charging mod](https://github.com/mralbobo/stardew-tool-charging) does this with both the watering can and hoe, though it'd be nice to see this implemented in Fast Animations.

Pathoschild:
I think skipping charge levels is best handled by the [Tool Charging mod](https://github.com/mralbobo/stardew-tool-charging) (which is linked from the Fast Animations mod page now). Fast Animations only speeds up animations, it deliberately doesn't change any game mechanics like that.

PR #107, opened 2017-09-11
Farmer Handler for speeding up all farmers actions.

agentd00nut:
By speeding up all the available animations for the farmer ( including running, walking, and carrying ) it's possible to overall increase the speed of the game.

The intention is that this lets a player set "FarmerSpeed = 2", then use another mod to set the games tick rate from 14 to 7.  Essentially the game is now playing twice as fast without offering the player any advantages.  Basically instead of ~20 minutes of real time to finish a day this change allows a player to finish a day in ~10 minutes of real time.

**Issues**
I was unable to get the waterbucket animation to play faster.  It seems to work similar to the eating animation in that it spawns temporary animations of the water pouring and splashing on the ground.  I attempted to use a more generic method that iterated over the current locations temporary animations as an array and update them accordingly but it didn't work.

Before pulling it would be great if you were able to write a WaterBucketHandler to specifically speed up the animation of the water bucket since right now a player would actually be able to water HALF as many crops if they speed up the games internal tick rate.

**Potential weirdness**
Technically the game will be in a slightly weird state if a player decreases the tick rate from 14 to 7 while doubling their own players animations.  How far a player can travel on the world map will stay consistent to vanilla play... A player can reach just into town by 6:10 in vanilla and with this mod.

However NPC's all move at the same rate... This **could** cause some interesting side effects where NPC's are late to their destinations.

Enemies in the mines will also move at a normal rate, causing a rather unbalanced state of gameplay where a player could easily out maneuver enemies.

**Fixing Weirdness**
Instead of updating the farmer object if it were possible to rewrite how **all** animations get played by doubling their updates the desired effect of "speeding up" the game would be achieved for anything that is animated.  This is pure speculation though and is likely much more difficult than it seems.


Pathoschild:
Hi @agentd00nut. Thanks for the pull request! This is a pretty cool idea, but I think it's out of scope for this mod. I'd rather take some of the animations you added to `FarmerSpeed` and add separate support for those, making sure they don't cause any weirdness like you mentioned above. You're welcome to fork the mod if you'd like to pursue that vision in more detail, though. :)

agentd00nut:
Understandable.  Would you be able to add a handler for the water bucket that I could then pull into my fork?  It's the one thing i wasn't able to get to work on my own.

Thanks for your time.

Pathoschild:
Yep, I'm planning to support that in a later version. You can watch #62 to know when it's implemented.

PR #117, opened 2017-10-19
Updated German Translation for ChestsAnywhere and LookupAnything

d0x7:
**ChestsAnywhere**

In German 'Versteck die Kiste' sounds like hiding the chest from the farm (like if it's invisible or something).
'Ausblenden' sounds like hiding it from the list, which is what this actually does.

**LookupAnything**

The most translations are fixed but I cannot guarantee that everything is fixed.


Pathoschild:
Thanks! Merged into `develop` for the upcoming Chests Anywhere 1.12 and Lookup Anything 1.17.

PR #152, opened 2018-01-07
Update manifest.json

Greavox:
Version format is not readeable by SDVMM V3.1g (Mod Manager) SMAPI 2.3


Pathoschild:
That's a bug in SDVMM per discussion. :)

PR #183, opened 2018-02-11
Added new DataMap: NeedsWatering

irecinius:
new DataMap: NeedsWatering
 - Display any crop that isn't watered.

Created new option to remove Data<ap from current playing session from the possible maps.
Restarting game or going to title screen will reset the maps.(at least should, and you cannot remove the last map, for now)

Added config option for key used to remove current DataMap: Default to F3
Added translations for NeedsWatering

Exposed CurrentMap from DataMapOverlay, so I could detect which map was loaded to remove from array.

Updated SMAPI NuGet package to 2.0.2


irecinius:
Git bugs durring commit.. reopened pull request

PR #184, opened 2018-02-11
Added new DataMap: NeedsWatering

irecinius:
Added new DataMap: NeedsWatering 
- Display any crop that isn't watered.

Created new option to remove DataMap from current playing session from the possible maps. Restarting game or going to title screen will reset the maps.
Added config option for key used to remove datamap: Default to F3

### DIFF
--- a/DataMaps/DataMaps.csproj
+++ b/DataMaps/DataMaps.csproj
@@ -11,6 +11,8 @@
     <AssemblyName>DataMaps</AssemblyName>
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
+    <NuGetPackageImportStamp>
+    </NuGetPackageImportStamp>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
     <DebugSymbols>true</DebugSymbols>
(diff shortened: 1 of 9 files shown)


irecinius:
- Added new DataMap: FertilizeMap

Pathoschild:
@irecinius Thanks! The new data maps look fine overall.

I'm not sure about the option to toggle data maps temporarily though. I'd rather address that by (a) remembering which map the player had open previously, and (b) adding `config.json` options to disable unwanted maps. Any objections if I remove that part of the pull request for now?

Pathoschild:
Two minor points:

* Separating fertilizers by quality might be confusing:  
    
  I think showing one option per fertilizer type (e.g. "Speed-Gro") would be simpler and sufficient. Most players probably won't need to differentiate by quality.
* For the _Needs Watering_ crop, showing a green 'Watered Crop' overlay might be useful too:  
  

If you have no objections, I can make those changes before merging the pull request.

Pathoschild:
Merged into `develop` for the upcoming release, with the changes discussed above. Thanks!

irecinius:
No problem with all changes, I'm for simplifying it.. I did give me a little headache to make all the colors work somehow..
Knowing the type only is a very good solution.

Going with the config.json route is very nice, I'd like to have done that way. I had that idea in mind in first place.

Glad I could help somewhat!

Pathoschild:
Thanks for the idea! I extended your config option into separate fields so it's more intuitive. Here's the new default config:

```json
{
  "CombineOverlappingBorders": true,
  "EnableMaps": {
    "Accessibility": true,
    "CoverageForBeeHouses": true,
    "CoverageForJunimoHuts": true,
    "CoverageForScarecrows": true,
    "CoverageForSprinklers": true,
    "CropFertilizer": true,
    "CropWater": true
  },
  "Controls": {
    "ToggleMap": "F2",
    "PrevMap": "LeftControl, LeftShoulder",
    "NextMap": "RightControl, RightShoulder"
  }
}
```

PR #193, opened 2018-02-28
Proposed Changes for issue 189

Fox536:
Ok so here's what I wrote, let me know what you think.

### DIFF
--- a/TractorMod/Framework/Attachments/AxeAttachment.cs
+++ b/TractorMod/Framework/Attachments/AxeAttachment.cs
@@ -19,9 +19,19 @@ namespace Pathoschild.Stardew.TractorMod.Framework.Attachments
 
         /// <summary>Whether to cut down fruit trees.</summary>
         private readonly bool CutFruitTrees;
+        
+        /// <summary>Whether to clear dead crops.</summary>
+        private readonly bool ClearDeadCrops;
 
         /// <summary>Whether to clear live crops.</summary>
-        private readonly bool ClearCrops;
+        private readonly bool ClearLiveCrops;
+        
+        /// <summary>Whether to clear debris.</summary>
+        private readonly bool ClearDebris;
+
+        /// <summary>Whether to clear tapped trees.</summary>
+        private readonly bool CutTappedTrees;
+
 
         /// <summary>The axe upgrade levels needed to break supported resource clumps.</summary>
         /// <remarks>Derived from <see cref="ResourceClump.performToolAction"/>.</remarks>
@@ -39,9 +49,14 @@ namespace Pathoschild.Stardew.TractorMod.Framework.Attachments
         /// <param name="config">The mod configuration.</param>
         public AxeAttachment(ModConfig config)
         {
-            this.CutTrees = config.AxeCutsTrees;
-            this.CutFruitTrees = config.AxeCutsFruitTrees;
-            this.ClearCrops = config.AxeClearsCrops;
+            this.ClearDeadCrops = config.StandardAttachments.Axe.ClearDeadCrops;
+            this.ClearLiveCrops = config.StandardAttachments.Axe.ClearLiveCrops;
+
+            this.CutFruitTrees = config.StandardAttachments.Axe.CutFruitTrees;
+            this.CutTrees = config.StandardAttachments.Axe.CutTrees;
+            this.CutTappedTrees = config.StandardAttachments.Axe.CutTappedTrees;
+
+            this.ClearDebris = config.StandardAttachments.Axe.ClearDebris;
         }
 
         /// <summary>Get whether the tool is currently enabled.</summary>
@@ -66,29 +81,37 @@ namespace Pathoschild.Stardew.TractorMod.Framework.Attachments
         {
             // clear twigs & weeds
             if (tileObj?.Name == "Twig" || tileObj?.Name.ToLower().Contains("weed") == true)
-                return this.UseToolOnTile(tool, tile);
+                return this.ClearDebris && this.UseToolOnTile(tool, tile);
 
             // cut non-fruit tree
             if (tileFeature is Tree)
-                return this.CutTrees && this.UseToolOnTile(tool, tile);
+            {
+                if ((tileFeature as Tree).tapped)
+                    return this.CutTappedTrees && this.UseToolOnTile(tool, tile);
+                else
+                    return this.CutTrees && this.UseToolOnTile(tool, tile);
+            }
 
             // cut fruit tree
             if (tileFeature is FruitTree)
                 return this.CutFruitTrees && this.UseToolOnTile(tool, tile);
 
-            // clear crops
-            if (tileFeature is HoeDirt dirt && dirt.crop != null && (dirt.crop.dead || this.ClearCrops))
-                return this.UseToolOnTile(tool, tile);
+            // clear dead crops
+            if (tileFeature is HoeDirt dirt && dirt.crop != null)
+                if (dirt.crop.dead && this.ClearDeadCrops)
+                    return this.UseToolOnTile(tool, tile);
+                else if (!dirt.crop.dead && this.ClearLiveCrops)
+                    return this.UseToolOnTile(tool, tile);
 
             // clear stumps
             // This needs to check if the axe upgrade level is high enough first, to avoid spamming
             // 'need to upgrade your tool' messages. Based on ResourceClump.performToolAction.
             {
                 ResourceClump clump = this.GetResourceClumpCoveringTile(location, tile);
-                if (clump != null && this.ResourceUpgradeLevelsNeeded.ContainsKey(clump.parentSheetIndex) && tool.upgradeLevel >= this.ResourceUpgradeLevelsNeeded[clump.parentSheetIndex])
+                if (clump != null && this.ResourceUpgradeLevelsNeeded.ContainsKey(clump.parentSheetIndex) && tool.upgradeLevel >= this.ResourceUpgradeLevelsNeeded[clump.parentSheetIndex] && this.ClearDebris)
                     this.UseToolOnTile(tool, tile);
             }
-
+            
             return false;
         }
     }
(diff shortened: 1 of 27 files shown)


TractorMod/Framework/ModConfig.cs
```diff
@@ -67,5 +55,141 @@ internal class ModConfigControls
             [JsonConverter(typeof(StringEnumArrayConverter))]
             public SButton[] HoldToActivate { get; set; } = new SButton[0];
         }
+
+        /// <summary>A set of control bindings.</summary>
+        internal class StandardAttachmentSet
```
Pathoschild: Each class here should be a separate file, probably in a `Framework/Config` folder and namespace. I also suggest renaming these to `StandardAttachmentsConfig`, `AxeConfig`, `ScytheConfig`, etc to avoid ambiguity with the attachment classes.

TractorMod/Framework/Attachments/ScytheAttachment.cs
```diff
@@ -10,9 +11,46 @@ namespace Pathoschild.Stardew.TractorMod.Framework.Attachments
...
+
+        /// <summary>Whether to harvest debris.</summary>
+        private readonly bool ClearDebris;
+
         /*********
         ** Public methods
         *********/
+        /// <summary>Construct an instance.</summary>
+        /// <param name="config">The mod configuration.</param>
+        public ScytheAttachment(ModConfig config)
```
Pathoschild: Since we have a separate config class for each attachment now, how about passing `ScytheConfig` into the constructor and saving a single `ScytheConfig` field (and doing the same for the other attachments)?

TractorMod/Framework/Attachments/AxeAttachment.cs
```diff
@@ -66,29 +81,37 @@ public override bool Apply(Vector2 tile, SObject tileObj, TerrainFeature tileFea
         {
             // clear twigs & weeds
             if (tileObj?.Name == "Twig" || tileObj?.Name.ToLower().Contains("weed") == true)
-                return this.UseToolOnTile(tool, tile);
+                return this.ClearDebris && this.UseToolOnTile(tool, tile);
```
Pathoschild: Ideally the new check should be part of the preconditions:
```cs
if (this.ClearDebris && (tileObj?.Name == "Twig" || tileObj?.Name.ToLower().Contains("weed") == true))
    return this.UseToolOnTile(tool, tile);
```

(I realise I did that myself further down. Feel free to correct that too if you want.)

TractorMod/Framework/Attachments/AxeAttachment.cs
```diff
@@ -66,29 +81,37 @@ public override bool Apply(Vector2 tile, SObject tileObj, TerrainFeature tileFea
...
 
             // cut non-fruit tree
             if (tileFeature is Tree)
-                return this.CutTrees && this.UseToolOnTile(tool, tile);
+            {
+                if ((tileFeature as Tree).tapped)
```
Pathoschild: We can simplify a bit from this:
```cs
if (tileFeature is Tree)
{
   if ((tileFeature as Tree).tapped)
      ...
}
```
...to this:
```cs
if (tileFeature is Tree tree)
{
   if (tree.tapped)
      ...
}
```

TractorMod/Framework/Attachments/AxeAttachment.cs
```diff
@@ -66,29 +81,37 @@ public override bool Apply(Vector2 tile, SObject tileObj, TerrainFeature tileFea
...
+                else if (!dirt.crop.dead && this.ClearLiveCrops)
+                    return this.UseToolOnTile(tool, tile);
 
             // clear stumps
             // This needs to check if the axe upgrade level is high enough first, to avoid spamming
             // 'need to upgrade your tool' messages. Based on ResourceClump.performToolAction.
             {
                 ResourceClump clump = this.GetResourceClumpCoveringTile(location, tile);
-                if (clump != null && this.ResourceUpgradeLevelsNeeded.ContainsKey(clump.parentSheetIndex) && tool.upgradeLevel >= this.ResourceUpgradeLevelsNeeded[clump.parentSheetIndex])
+                if (clump != null && this.ResourceUpgradeLevelsNeeded.ContainsKey(clump.parentSheetIndex) && tool.upgradeLevel >= this.ResourceUpgradeLevelsNeeded[clump.parentSheetIndex] && this.ClearDebris)
```
Pathoschild: Ideally check the config flags first (e.g. `this.ClearDebris && <other conditions here>` instead). Checking a bool field is quicker than the other preconditions (e.g. two dictionary lookups), so we can avoid some unnecessary work that way.

TractorMod/Framework/Attachments/AxeAttachment.cs
```diff
@@ -66,29 +81,37 @@ public override bool Apply(Vector2 tile, SObject tileObj, TerrainFeature tileFea
...
 
             // cut fruit tree
             if (tileFeature is FruitTree)
                 return this.CutFruitTrees && this.UseToolOnTile(tool, tile);
 
-            // clear crops
-            if (tileFeature is HoeDirt dirt && dirt.crop != null && (dirt.crop.dead || this.ClearCrops))
-                return this.UseToolOnTile(tool, tile);
+            // clear dead crops
+            if (tileFeature is HoeDirt dirt && dirt.crop != null)
```
Pathoschild: The brackets are missing here, so only the immediate next block is inside the outer `if`. The new code is equivalent to this:
```cs
if (tileFeature is HoeDirt dirt && dirt.crop != null)
{
   if (dirt.crop.dead && this.ClearDeadCrops)
      return this.UseToolOnTile(tool, tile);
}
else if (!dirt.crop.dead && this.ClearLiveCrops)
   return this.UseToolOnTile(tool, tile);
```

To avoid confusion, I add brackets anytime there's more than one statement within.

TractorMod/Framework/Attachments/PickaxeAttachment.cs
```diff
@@ -77,15 +85,15 @@ public override bool Apply(Vector2 tile, SObject tileObj, TerrainFeature tileFea
...
+                    return this.ClearDeadCrops && this.UseToolOnTile(tool, tile);
             }
 
             // clear boulders / meteorites
             // This needs to check if the axe upgrade level is high enough first, to avoid spamming
             // 'need to upgrade your tool' messages. Based on ResourceClump.performToolAction.
             {
                 ResourceClump clump = this.GetResourceClumpCoveringTile(location, tile);
-                if (clump != null && this.ResourceUpgradeLevelsNeeded.ContainsKey(clump.parentSheetIndex) && tool.upgradeLevel >= this.ResourceUpgradeLevelsNeeded[clump.parentSheetIndex])
+                if (clump != null && this.ResourceUpgradeLevelsNeeded.ContainsKey(clump.parentSheetIndex) && tool.upgradeLevel >= this.ResourceUpgradeLevelsNeeded[clump.parentSheetIndex] && this.ClearDebris)
```
Pathoschild: A separate config field for clearing boulders / meteorites might be useful here.

TractorMod/Framework/Attachments/ScytheAttachment.cs
```diff
@@ -46,27 +84,46 @@ public override bool Apply(Vector2 tile, SObject tileObj, TerrainFeature tileFea
...
                     return false;
 
-                if (dirt.crop.dead)
+                if (dirt.crop.dead && this.ClearDeadCrops)
                 {
                     this.UseToolOnTile(new Pickaxe(), tile); // clear dead crop
                     return true;
                 }
+
+                List<int> flowers = new List<int> { 376, 591, 593, 595, 597, 421 };
```
Pathoschild: Let's take out the flower feature for now, and create a separate pull request for it. I'd want to change how we get these IDs (maybe add a `data.json` to support players with custom flowers). For this pull request we can use `this.HarvestCrops` for flowers too.

TractorMod/Framework/Attachments/ScytheAttachment.cs
```diff
@@ -46,27 +84,46 @@ public override bool Apply(Vector2 tile, SObject tileObj, TerrainFeature tileFea
...
+                {
+                    if (flowers.Contains(dirt.crop.indexOfHarvest))
+                        return this.HarvestFlowers && dirt.performToolAction(tool, 0, tile, location);
+                    else
+                        return this.HarvestCrops && dirt.performToolAction(tool, 0, tile, location);
+                }
+                else
+                    this.CheckTileAction(location, tile, player);
+
+                /*
```
Pathoschild: No need to comment out code, since it's all in source control. Feel free to just remove old code.

Pathoschild (changes requested):
Thanks for the contribution!

I suggested a few changes below. Let me know if you'd prefer I make the changes myself when I accept the pull request; I realise I can be a bit picky. :)

TractorMod/Framework/Attachments/AxeAttachment.cs
Fox536: Huh, I didn't catch that I thought it used following else with the inner most if, good to know, and I really should of added brackets either way it makes it easier to read.

TractorMod/Framework/Attachments/PickaxeAttachment.cs
Fox536: You think one or two settings here, I'm thinking one but I can do either.

TractorMod/Framework/Attachments/ScytheAttachment.cs
Fox536: Good, that one was tricky and I wasn't really happy with the results. I guess there isn't a flag anywhere to check if a plant is a flower or not, which would be helpful.

TractorMod/Framework/Attachments/ScytheAttachment.cs
Fox536: Huh, I thought I removed it once I was finished, I guess I missed that section though.

Fox536:
Ok, updated the code fixed the changes, let me know if you need anything else changed. 

For the flowers I removed it for now, I could add an array to the ScytheConfig to check against. That way if people have custom flowers they can add the Ids from it to the array as they need to.

Pathoschild:
Thanks! I'll keep this open until the next Tractor Mod release and merge it in then.

Pathoschild (approved):

Pathoschild:
Merged into `develop` for the next release. Thanks!

PR #199, opened 2018-03-02
Enabled flag for entire Content pack

f4iTh:
The feature may not be useful for the _average player_, but it might be a life saver for some modders. Being able to disable a content pack without having to remove any of the files can be convenient.

What would be a nice addition to this however, would be for SMAPI to read through the _content.json_ files of each Content pack first so it wouldn't load the ones that will be disabled for nothing. 
screenshot_531
A simple photo example. 

This would be a nice addition but I don't think it's exactly _needed_. 
Useful? Yes. Must have? Not really.


Pathoschild:
Thanks for the pull request!

I'm not sure we need an `Enabled` flag for the whole content pack though. I want to keep the JSON file as simple as possible, so it's easy for modders to learn. Disabling the entire content pack seems to me like something that should be handled by a mod manager, SMAPI (e.g. an `Enabled` field in `manifest.json`), or the player (zip and delete the folder).

PR #254, opened 2018-05-12
[LookupAnything] Bug fixed & display text added

StefanOssendorf:
As in discord discussed, here my PR based on stable (tested in 1.2).

Bug: If an item is used multiple times (e.g. wood) the first stack would make the lookup say it's not necessary anymore.

Improvement: If an item needs more than 1 it's now displayed in the lookup


Pathoschild:
Thanks! Merged into `develop` for the upcoming Lookup Anything 1.19-beta.3.

PR #270, opened 2018-05-20
[Fast Animations] speed up game load delay

danvolchek:
This pull request adds support for fast animations to shorten the delay that occurs before the game actually starts loading when the player clicks on a save slot.

My naming of the config variable/description is not ideal, so I would appreciate advice on a better way to describe what is being sped up.

Please let me know of any changes I should make :)


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Fast Animations 1.6-beta.3 release. Thanks!

PR #285, opened 2018-06-16
Updated Chinese translation

changbowen:
Do you have an official Chinese name for the Automate mod?

### DIFF
--- a/ChestsAnywhere/i18n/zh.json
+++ b/ChestsAnywhere/i18n/zh.json
@@ -13,8 +13,8 @@
   "label.order": "顺序",
   "label.hide-chest": "隐藏此箱子",
   "label.hide-chest-hidden": "隐藏此箱子（你需要亲自找到该箱子撤销隐藏！）",
-  "label.automate-prefer-output": "(Automate) Put items in this chest first", // TODO
-  "label.automate-ignore": "(Automate) Don't use this chest for automation", // TODO
+  "label.automate-prefer-output": "(Automate) 优先输出到这个箱子",
+  "label.automate-ignore": "(Automate) 忽略这个箱子",
 
   // button labels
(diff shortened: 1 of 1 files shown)


Pathoschild:
@changbowen Thanks! Nope, there's no official Chinese name. Maybe a translation for "Automate mod" would be clear enough in this context?

Pathoschild:
Merged into `develop` for the upcoming Chests Anywhere 1.13-beta.6. Thanks!

## Other missing translations for Chinese
If you're interested, there are some missing translations for [Debug Mode](https://www.nexusmods.com/stardewvalley/mods/679). It'd be great if you could submit a pull request for those too. :)
```js
{
  // labels shown on the debug screen
  "label.tile": "地格",
  "label.map": "地图",
  "label.menu": "菜单",
  "label.submenu": "submenu", // TODO
  "label.festival-name": "festival", // TODO
  "label.event-id": "event ID", // TODO
  "label.event-script": "event script" // TODO
}
```
Here's where they appear:
> menu info
> event info
> festival info

changbowen:
If no official Chinese name, we can leave it untranslated so it doesn't cause misunderstandings.
About Debug Mode translation, I was not very clear about what the item "event script" means. Based on the screenshot it looks like the name of the script running for the event. Because script can both mean script as in scripting language, or script as in a screenplay... and it's different Chinese words :)

Pathoschild:
@changbowen An event has many commands to add dialogue, move characters, etc. The 'event script' field shows each command that's run. For example, the screenshot shows the current command as `emote Marnie 32` which adds the owo emote above Marnie's head. So it's kind of a screenplay script, but feel free to translate it differently if needed (like "current command").

changbowen:
Thanks for the explanation. In this case i think it makes sense to stick to the current translation :)

PR #290, opened 2018-06-24
[TheLongNight] Prevent pass out by freezing time before 2am

mattfeldman:
I simplified the mechanism to prevent pass out by freezing time at 1:50am via `Game1.gameTimeInterval`, are there issues or limitations with this approach?

I've tested:
- In multiplayer
- Fishing unlocks at limit
- Machines pause at limit
- Game end -> save -> new day as expected

### DIFF
--- a/Pathoschild.Stardew.sln
+++ b/Pathoschild.Stardew.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.27130.2036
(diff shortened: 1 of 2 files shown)


Pathoschild:
Thanks for the pull request!

The Long Night is really about staying up past 2am while time continues to run (e.g. for machine processing). Unfortunately freezing time doesn't really fit the mod's goals, and there are already mods to do that (e.g. TimeSpeed can freeze time once it reaches a specified time).

I'm considering whether to use Harmony to reimplement the mod though, since the game encapsulated the collapse logic into a method we could patch in 1.3.21.

Issue #151, opened 2018-01-06
[Chest Anywhere] Browse through chests by mouse scrolling?


changbowen:
Is it possible to use the mouse scroll wheel to browse through the chests?

Pathoschild:
@changbowen You can navigate quickly using the arrow keys on the keyboard. Is that sufficient?

changbowen:
@Pathoschild is it supported by the framework to use mouse wheel up / down instead of the key press?

Pathoschild:
Yep, SMAPI does support handling mouse scroll.

changbowen:
what keybinding value to use in the config.json file?
i cannot find values for mouse wheel action in https://stardewvalleywiki.com/Modding:Key_bindings

Pathoschild:
SMAPI has an event for mouse changes, but the mod needs to be programmed to support it. That's why I was asking if you knew about the keyboard navigation — if that was sufficient, there'd be no reason to look into support for mouse wheel scrolling. :)

changbowen:
it would be a nice addition for me :)
but if it is too much trouble i can live with the keyboard navigation too.

Pathoschild:
Done via #291 for the upcoming Chests Anywhere 1.13-beta.7 (thanks to @mattfeldman!). This adds two shortcuts:

keys | result
---- | ------
`Left Control` + scroll mouse wheel | scroll chest dropdown.
`Left Alt` + scroll mouse wheel | scroll category dropdown.

PR #291, opened 2018-06-24
[Chests Anywhere] Support scrolling through chests and categories

mattfeldman:
For #151 

This lets you scroll through chests by default and categories while holding a modifier key. The modifier is configurable (shift by default) and the scrolling support can be disabled as well.


Pathoschild:
I made a few changes:
* Replaced the `ScrollModifierController` with SMAPI's new input API.
* Changed to more explicit modifier keys:

   keys | result
   ---- | ------
   `Left Control` + scroll mouse wheel | scroll chest dropdown.
   `Left Alt` + scroll mouse wheel | scroll category dropdown.

* Removed the `EnableScrollNavigation` setting (can disable it by removing the modifier key bindings).

Pathoschild:
Thanks! Merged into `develop` for the upcoming Chests Anywhere 1.13-beta.7.

PR #297, opened 2018-07-06
Portuguese translation of Data Layers

TaelFayre:
ps: i'm not gonna use the mod, but it sounds really useful so if somebody needs it in portuguese, the translation will be ready.


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Data Layers 1.4-beta.5. Thanks!

PR #359, opened 2019-01-05
Update German translations for Lookup Anything

Nanogamer7:
I'm not quite sure if pull requests are still the right way to submit translations, but from my knowledge they are pretty easy to handle, so here it goes anyways


Pathoschild (approved):

Pathoschild:
Thanks! Pull requests are still the best way to submit translations. Merged into `develop` for the next Lookup Anything release.

PR #369, opened 2019-01-27
Enable content packs for Tractor Mod

paritee:
Scenario: Modders have to manually overwrite TractorMod's assets if they want to change the look of the tractor and garage

- Added a check OnGameLaunched for content packs. Content packs should follow the format Mods/ContentPack/tractor.png or Mods/ContentPack/garage.png (could also use the Mods/ContentPack/assets/spritesheet.png structure if you're planning on doing more with content packs in the future)
- Defaults to TractorMod's assets if none could be found in the content packs. If there are multiple content packs for the same asset, the first asset found will be used
- Refactored season formatting into its own function for reusability

### DIFF
--- a/TractorMod/ModEntry.cs
+++ b/TractorMod/ModEntry.cs
@@ -71,6 +71,9 @@ namespace Pathoschild.Stardew.TractorMod
         /// <summary>Whether the mod is enabled for the current farmhand.</summary>
         private bool IsEnabled = true;
 
+        /// <summary>Content packs loaded for TractorMod.</summary>
+        private IEnumerable<IContentPack> ContentPacks;
+
 
         /*********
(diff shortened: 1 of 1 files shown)


Pathoschild:
Thanks for the pull request!

Content pack support is something I'm looking into, but I want to support multiple tractor skins (e.g. so each player can have a different tractor in multiplayer). That adds a few complications like syncing tractor skins in multiplayer, deciding how to assign tractor skins, etc.

This implementation would work fine in single-player though, so I might merge this in for the next release and look into multi-skin support in a future version.

Pathoschild:
I'll close this PR since it's stale, but I may still use the changes in a future implementation per my last comment. Thanks again!

PR #401, opened 2019-04-08
Change default tractor distance

Kirin224:


Pathoschild:
Thanks for the pull request! Unfortunately that default value would negatively impact performance and usability. You're welcome to use the value in your own `config.json` though.

PR #422, opened 2019-05-19
Basic API 2: Rebase boogaloo

spacechase0:
(Okay, a rebase was too much work, so I just cherry-picked.)

See #420 for details


Pathoschild (approved):

Pathoschild:
Thanks! Merged manually due to conflicting changes, for the upcoming Content Patcher 1.9.

I also rewrote the implementation so mods don't need to reference `ContentPatcher.dll` directly, and to add lots of validation to avoid edge cases (e.g. if a content pack references a token but doesn't list the mod that provided it as a dependency).

PR #448, opened 2019-08-15
Update Russian translations for Lookup Anything

ghost:
Fixed incorrectly translated words


Pathoschild (approved):

Pathoschild:
Thanks for the help! I added it to the next Lookup Anything release.

I have [two other mods which need Russian translations](https://github.com/Pathoschild/StardewMods#translating-the-mods) if you're interested. (They only have a few lines to translate, unlike Lookup Anything.)

PR #465, opened 2019-11-06
Add token HasDialogueQuestionAnswered

mus-candidus:
I wanted to do things in ContentPatcher depending on certain responses given to questions. Your code structure made it really easy to do that. I'm still unsure about the naming of the token so I used the corresponding field name, maybe there are better options.


Pathoschild (approved):

Pathoschild:
Thanks! Manually merged into `develop` for the upcoming Content Patcher 1.10.

PR #466, opened 2019-11-12
Generic Mod Config Menu support for Content Patcher

spacechase0:
I wasn't sure about the keybindings (since you allow multiple at once), so I just made it so that only the first one can be configured.

GMCM still hasn't been officially released yet, so let you know if you want any changes to better support things!

### DIFF
--- a/Common/Common.projitems
+++ b/Common/Common.projitems
@@ -13,6 +13,8 @@
     <Compile Include="$(MSBuildThisFileDirectory)Integrations\BaseIntegration.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Integrations\Automate\AutomateIntegration.cs" />
     <Compile Include="$(MSBuildThisFileDirectory)Integrations\Automate\IAutomateApi.cs" />
+    <Compile Include="$(MSBuildThisFileDirectory)Integrations\GenericModConfigMenu\GenericModConfigMenuIntegration.cs" />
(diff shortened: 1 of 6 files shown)


Pathoschild:
I'm mostly offline temporarily while I start my new job, but I took a quick look. A few comments:

* `ManualToken` can change value between context updates, which may lead to subtle/strange bugs. For example, it may return a different value after it was already validated (e.g. CP checks if the file exists, then the file being referenced changes). The simplest approach short of refactoring Content Patcher would be to cache the new value, and apply it next time the token receives a context update.
* The API has a lot of very specific methods. A fluent interface would be nice, but it's not really workable without a hard reference. Maybe you could have a few generic methods with optional arguments like this?
   ```cs
   RegisterOption<T>(
      string name,
      string description,
      Func<T> get,
      Action<T> set,
      T? min = null,
      T? max = null,
      T[] choices = null
   );
   ```

Pathoschild:
Looking into this a bit closer:

* It introduces a major breaking change. Since config values are no longer immutable, (a) some of Content Patcher's internal assumptions about how to optimise them break, and (b) content packs which use config tokens in an immutable context like the `Enabled` field will no longer be allowed to do so.

  Fixing that would probably require rewriting parts of Content Patcher. With Stardew Valley 1.4 releasing soon, I'll keep this out of the coming release so we can figure out how to integrate it without breaking existing content packs.
* It shouldn't trigger a context update whenever a content pack's config changes. (That one should be a simple change though.)

Pathoschild:
Unfortunately I have no current plans to tackle this due to the issues mentioned above, sorry!

PR #480, opened 2019-12-05
Color blending for EditImage

mus-candidus:
Hi, I'm working a lot with overlays in ContentPatcher so I added color blending. It's simple multiplicative blending and can be controlled by the parameter "BlendColor" in config.json.
The attached (unpublished) example mod uses the feature to change Abigail's hair color.

[SilverFoxAbigail.zip](https://github.com/Pathoschild/StardewMods/files/3929313/SilverFoxAbigail.zip)


Pathoschild:
Thanks for the PR! I'm not opposed to the feature, but could you list some possible applications? For example, it seems recolor modders mostly need to make manual edits so they probably wouldn't use it.

mus-candidus:
I'm thinking of configurable recolor mods: You can use a grayscale overlay and multiply it with a color to recolor hair, clothes or whatever you want. Maybe we should ask other modders if they'd appreciate that feature to make sure I'm not the only one:-) Is the discord channel the right place or is there another place to discuss feature requests?

Pathoschild:
Yep, Discord is a good place to discuss with other modders.

mus-candidus:
Well, I'm really bad at advertising but at least one other modder seems to be interested. So that makes two :-)

Pathoschild:
Unfortunately this is a bit too niche to add to Content Patcher itself, sorry! I'll look into letting mods add custom patch types through the API though.

Issue #494, opened 2019-12-30
[Data Layers] Allow key bindings for specific layers


Drachenkaetzchen:
It would be extremely useful to have key bindings when data layers are active. For example, map the key "1" to "Crops: Ready to harvest" and "2" to "Crops: Watered".

Pathoschild:
Thanks! Added via #504.

PR #504, opened 2020-01-04
Added shortcut buttons for direct layer activation

Drachenkaetzchen:
Implementation for #494


Drachenkaetzchen:
I'm not good with wording for the README, so here are some verbose descriptions.

Assume the `TillableLayer` shortcut is set to `D1`, `ToggleLayer`to `F2`.

If `ActivateOverlayOnShortcut` is set to false, the user first has to activate the overlay via the configured `ToggleLayer` button (`F2` then `D1`).

If `ActivateOverlayOnShortcut` is set to true, the shortcut automatically activates the overlay (`D1`). 

If `ToggleOverlayViaShortcut`, the overlay is toggled when using shortcuts. First `D1` enables the overlay and shows the `TillableLayer`, another `D1` disables the overlay again.

Pathoschild:
@Drachenkaetzchen To keep usage straightforward, I simplified that a bit so the layer key only applies while the overlay is open. We could reintroduce the extra options later if it's something a lot of players want though.

Pathoschild (approved):

Pathoschild:
Thanks! Merged into `develop` for the next Data Layers update.

PR #521, opened 2020-02-14
[Automate] Add support for tea bushes

stellarashes:
This PR adds support for tea bushes.  Some of the state checking comes from [BushSubject](https://github.com/Pathoschild/StardewMods/blob/develop/LookupAnything/Framework/Subjects/BushSubject.cs).  I'm not entirely sure about the best way to "reset" the bush's state after taking output, but in testing using `performUseAction` seems to work okay.

### DIFF
--- a/Automate/Framework/AutomationFactory.cs
+++ b/Automate/Framework/AutomationFactory.cs
@@ -162,7 +162,7 @@ namespace Pathoschild.Stardew.Automate.Framework
             // machine
             if (feature is FruitTree fruitTree)
                 return new FruitTreeMachine(fruitTree, location, tile);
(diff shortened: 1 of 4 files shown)


stellarashes:
OK did some digging around and I found the source for shake in SDV.  There was some problem with the original implementation as plopping down a chest next to a tea leaf would both yield 1 in chest + shake the bush to drop one.  The updated one will no longer do that and work as expected.  Also added  a check for town bush.

Pathoschild (approved):

Pathoschild:
Thanks! Merged into `develop` for the next Automate update.

PR #522, opened 2020-02-18
[Automate] add compatibility with Mega Storage

LeFauxMatt:
Tested with and without MegaStorage and in both cases the expected behavior is occuring.


LeFauxMatt:
Mega Storage uses Capacity since it cannot override the default Chest capacity constant. This enables Automate to handle Mega Storage's capacity as well as default chest.

Pathoschild (approved):

Pathoschild:
Thanks! Merged into `develop` for the next Automate update.

PR #523, opened 2020-02-21
[Content Patcher] add missing layers and properties in map patches

notevenamouse:
Change ApplyMapOverride to add new layers if needed, and copy layer properties if present. Totally not stolen from PyMaps.


Pathoschild (approved):

Pathoschild:
Thanks! Merged into `develop` for the upcoming Content Patcher 1.13.

PR #525, opened 2020-03-01
Fix compatibility between tractor mod & harvest with scythe mod

bcmpinc:
This fixes the weird incompatibility issue between those two mods.

The problem is that Harvest with scythe slightly modifies the meaning of `Crop.harvestMethod.Value`. This results in the tractor mod to call `HoeDirt.performUseAction` on a crop that subsequently is considered a `Crop.sickleHarvest` crop causing it to trigger a harvest with scythe animation. By calling `dirt.crop.harvest` directly this problem is circumvented.


Pathoschild (approved):

Pathoschild:
Thanks! Added in the upcoming Tractor Mod 4.10.

PR #542, opened 2020-04-02
[Fast Animations] Add support for non-integral update rates.

### LINKED ISSUE #541: [Fast Animations] Support fractional playback rates
I am finding that 200% is a little too fast for tool and weapon usage.  However, specifying a real number (e.g. 1.5 for 150% speed) results in an error parsing the configuration as the mod expects an integer.

Would it be possible to support non-integral values to fine-tune the playback rate?

**Edit:** Taking a quick look at the code, it appears this mod works by calling `Update` additional times per normal time step.  It makes sense why an integer value is used.  To support a rational value, it should only be necessary to keep an internal counter that increments by the requested rate each time step.  This would mean for 1.5x, the extra `Update` happens every other normal time step.  When I get a local build environment running, I will try out my approach and submit a PR should it work.

seraku24:
Fixes #541.

Verified the new logic works by recording gameplay and counting frames of animation of a weapon swing (from the first appearance of the sword to the character resuming a standing pose).

|Rate|Frame Count|
|---|---|
|1.0|24|
|1.5|16|
|2.0|12|

NOTE: Your coding conventions differ from mine for C#, so I hope I have matched your style properly.  Let me know if I need to correct anything.

### DIFF
--- a/FastAnimations/Framework/BaseAnimationHandler.cs
+++ b/FastAnimations/Framework/BaseAnimationHandler.cs
@@ -8,10 +8,16 @@ namespace Pathoschild.Stardew.FastAnimations.Framework
     internal abstract class BaseAnimationHandler : IAnimationHandler
     {
         /*********
+        ** Fields
+        *********/
+        /// <summary>Tracks the number of updates.</summary>
+        private float UpdateCounter = 0;
+
+        /*********
         ** Accessors
         *********/
         /// <summary>The animation speed multiplier to apply.</summary>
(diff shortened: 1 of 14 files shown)


Pathoschild (approved):

Pathoschild:
Thanks for the PR! Merged into `develop` for the upcoming Fast Animations 1.8.

Some animations don't use `SpeedUpPlayer`, but I'll fix those after the merge.

PR #543, opened 2020-04-08
Android compatibility for SkipIntro

kdau:
These changes produce effectively the same behavior on Android as the mod has for me on Linux. The lighter touch here is sufficient to skip the reduced animations on the Android port of TitleMenu. The black screen issue was due to the `buttonsToShow` loop never finishing. Skipping again on return to title is unnecessary for Android and traps the player in a submenu when SkipTo != Title.


Pathoschild (approved):

Pathoschild:
Thanks! Merged into `develop` for the upcoming Skip Intro 1.8.4.

PR #545, opened 2020-04-17
Update Chinese translations for Lookup Anything

niniack:


Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Lookup Anything 1.26.1.

niniack:
Happy to contribute!

Referring to the remaining couple needed translations, (lines 557 & 558) I was a bit hesitant to touch them because I'm not quite sure ix means. Any suggestions?

```
    "tile.index-property": "{{layerName}}: ix props: {{propertyName}}", // TODO
    "tile.tile-property": "{{layerName}}: props: {{propertyName}}", // TODO
```

Pathoschild:
@niniack Those are properties stored in the map file. They have values like `Diggable` (which tells the game you can dig there), `Type` (which changes which footstep sounds play), etc. There are two tile property types:
* An _index property_ (shortened to _ix prop_ in the English text) is linked to the tile ID in the spritesheet. For example, 175 is a grass tile with a `Type: Grass` index property, so every tile with that ID has that property.
* A _tile property_ is linked to a specific position on the map. For example, tile `(74, 17)` on the farm has an `Action: Mailbox` property.

Let me know if it's still unclear; I'll add an explanation in the translation files in the next update.

Issue #307, opened 2018-07-16
[Automate] add local storage indexes


Pathoschild:
Add 'local storage indexes' to significantly improve performance and enable efficient transport pipes.

Pathoschild:
## Proposed implementation
### Background
One of the difficulties for transport pipes (#61) and performance (#281) is efficiently finding items available to a machine. Hundreds of machines may need to perform multiple queries at once without causing visible stutter. Local storage indexes provide a way to do that.

### Current implementation
Each machine group has a `StorageManager` which encapsulates access to the containers within the group. To search or list items, it essentially iterates through every item slot in every container. We can boil that down to this diagram (ignoring a few abstractions that don't matter here):

```
┌───────┐
│ chest │<──┐ find                       find
╞═══════╡   │ item X ┌─────────────────┐ item X ┌─────────┐
│ chest │<──┼────────┤ storage manager │<───────┤ machine │
╞═══════╡   │        └─────────────────┘        └─────────┘
│ chest │<──┘
└───────┘
```

That can impact performance in some cases: many machines might each perform multiple searches, which each iterate multiple containers, which each iterate multiple item slots. The problem gets significantly worse with transport pipes, which increases the number of potential containers, adds filtering logic, and adds complications like multiple in/out connections and multiple routes between machine groups (not to mention potential circular references).

### Proposed implementation (local storage indexes)
A _local storage index_ is an optimised view of the connected containers. It contains references to every item stack in those containers, and stacks added/removed in the containers get added/removed in the index (without rebuilding the rest of the index). It doesn't need to be updated when a stack size changes, since it holds references to the stacks.

For example (showing 3 slots for simplicity, but each container can have 36+ slots):
```
┌─────────────┐
│ chest       │<──┐
│  [3 item X] │   │ get
│  [empty]    │   │ all
│  [empty]    │   │ items  ┌─────────────┐
╞═════════════╡   │        │ local index │
│ chest       │<──┼────────┤  [3 item X] │
│  [5 item Y] │   │        │  [5 item Y] │
│  [empty]    │   │        │  [1 item Z] │
│  [empty]    │   │        └─────────────┘
╞═════════════╡   │
│ chest       │   │
│  [1 item Z] │<──┘
│  [empty]    │
│  [empty]    │
└─────────────┘
```

With the index in place, the previous example is now much simpler. Instead of checking every slot in every container (108+ checks per search in the above example), we can search the optimised index (3 checks per search):
```
           find                       find
[...]

PR #546, opened 2020-04-18
[Automate] Add the option to always keep a certain amount of each item

paranoidduck:
New configuration option: "KeepAtLeast" (integer, defaults to 0).

When this option is set to a positive value, machines will always leave at least this amount in its associated chest group, regardless of the quality.

The lowest quality items are consumed first, meaning that those kept in storage are of the best quality available.


Pathoschild:
Thanks for the pull request!

Unfortunately this will noticeably impact performance in many cases. When a machine looks for input, the inventory for every connected chest will now be copied, sorted, and iterated multiple times; that will happen for every machine, potentially hundreds or thousands of times per second.

There are planned features (e.g. #307) that will make this type of querying more efficient, but in the meantime it's important for performance to keep the streaming-style approach, where the minimum number of items are iterated each time.

PR #551, opened 2020-05-28
Added a config menu for Tractor Mod on F2

NexusFlight:
I have added an in game menu so you are able to configure the tractor mod in game
This does not include price or materials of the garage as I felt that was unneeded
The values for Distance, Speed, Magnet Range scale as default, *4 *8 *16


Pathoschild:
Thanks for the PR, I appreciate the contribution! I wish you'd discussed it first though, since the ≈550 lines of custom UI code in a non-UI mod are the main reason I haven't added one yet. I'd prefer an optional integration with [Generic Mod Config Menu](https://www.nexusmods.com/stardewvalley/mods/5098) instead, so Tractor Mod only needs to handle the tractor-related logic.

NexusFlight:
That is more than understandable, I have plenty of spare time at the moment so I'll close this request and when I get round to making the integration I'll let you know.

NexusFlight:
I've been playing around with Generic Mod Config Menu 
And have got this to function with Tractor Mod, I wrote the integration Api in your Common Shared Project.
However the code that generates the menu very much like the creators(spacechase0) [example](https://gist.github.com/spacechase0/2d8d4dbffe5f2ce9457d2c891a8b99e3/)
is within the mod entry file.
I was wondering if that was your intention or to have a whole new mod that does this?
I didn't want to spam you with another Pull request if it didn't match how you wanted it to work.
image

Pathoschild:
Yep! I haven't used GMCM yet, but does it support sliders for the numeric values so we can set some reasonable min/max values?

PR #561, opened 2020-08-06
[Content Patcher] Allow unloading top-level patches

spacechase0:
Ability to unload a whole content pack.
Added `patch reload` command, to reload an entire content pack. (Patches only, config schema and dynamic tokens unsupported.) Mainly for testing this. I can take it out if you don't want it in for whatever reason (like not supporting the other two things.)

I made my VS2017 changes a separate commit so this PR's commit could easily be cherry-picked. I don't know enough about git to get just this commit onto the branch after the fact. :P

### DIFF
--- a/ContentPatcher/Framework/Commands/CommandHandler.cs
+++ b/ContentPatcher/Framework/Commands/CommandHandler.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
 using ContentPatcher.Framework.Conditions;
+using ContentPatcher.Framework.ConfigModels;
 using ContentPatcher.Framework.Lexing.LexTokens;
 using ContentPatcher.Framework.Patches;
 using ContentPatcher.Framework.Tokens;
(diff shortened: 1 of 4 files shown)


Common/CommonHelper.cs
```diff
@@ -418,7 +418,7 @@ public static void InterceptErrors(this IMonitor monitor, string verb, string de
         /// <param name="detailedVerb">A more detailed form of <see cref="verb"/> if applicable. This is displayed in the log, so it can be more technical and isn't constrained by the sprite font.</param>
         public static void InterceptError(this IMonitor monitor, Exception ex, string verb, string detailedVerb = null)
         {
-            detailedVerb ??= verb;
+            detailedVerb = detailedVerb ?? verb;
```
Pathoschild: Can you remove C# backwards-compatibility changes from the PR?

ContentPatcher/Framework/Patches/DisabledPatch.cs
```diff
@@ -27,7 +27,7 @@ internal class DisabledPatch
         public string ReasonDisabled { get; }
 
         /// <summary>The parent patch for which this patch was loaded, if any.</summary>
-        public Patch ParentPatch { get; }
+        public IPatchSource PatchSource{ get; }
```
Pathoschild: This is a pretty neat approach, but I think it'd be more flexible long-term to track both:
```cs
/// <summary>The parent patch for which this patch was loaded, if any.</summary>
public Patch ParentPatch { get; }

/// <summary>The content pack for which this patch was loaded.</summary>
public RawContentPack ContentPack { get; }
```
That also avoids casting checks like `(source as IPatch)?.PatchSource` and `source is IPatch parent` in the rest of the code.

Pathoschild (changes requested):
Thanks! A few comments:

Pathoschild (approved):

Pathoschild:
Thanks! Merged into `develop` for the next Content Patcher update.

PR #563, opened 2020-08-09
GMCM support for CP packs (not CP itself though)

spacechase0:
I didn't entirely understand the changes I needed to make to `ManualValueProvider`, so let me know if those are off.

Tested with Lumisteria Flowery Grass.

### DIFF
--- a/ContentPatcher/Framework/ModTokenContext.cs
+++ b/ContentPatcher/Framework/ModTokenContext.cs
@@ -155,7 +155,7 @@ namespace ContentPatcher.Framework
             {
                 foreach (ManagedManualToken managed in this.DynamicTokens.Values)
                 {
-                    managed.ValueProvider.SetValue(null);
+                    managed.ValueProvider.SetValue((ITokenString)null);
                     managed.ValueProvider.SetReady(false);
                 }
 
--- a/ContentPatcher/Framework/ModTokenContext.cs
+++ b/ContentPatcher/Framework/ModTokenContext.cs
@@ -75,6 +75,14 @@ namespace ContentPatcher.Framework
             this.HasNewTokens = true;
         }
 
+        /// <summary>Remove a registered local token.</summary>
+        /// <param name="name">The config token to remove.</param>
+        public void RemoveLocalToken(string name)
+        {
+            this.LocalContext.Remove(name);
+            this.HasNewTokens = true; // update dynamic tokens
+        }
+
         /// <summary>Add a dynamic token value to the context.</summary>
         /// <param name="name">The token name.</param>
         /// <param name="rawValue">The token value to set.</param>
@@ -155,7 +163,7 @@ namespace ContentPatcher.Framework
             {
                 foreach (ManagedManualToken managed in this.DynamicTokens.Values)
                 {
-                    managed.ValueProvider.SetValue((ITokenString)null);
+                    managed.ValueProvider.SetValue(null);
                     managed.ValueProvider.SetReady(false);
                 }
 
--- a/ContentPatcher/ModEntry.cs
+++ b/ContentPatcher/ModEntry.cs
@@ -523,24 +523,25 @@ namespace ContentPatcher
                 if (configField.AllowMultiple)
                 {
                     // Whitelist + multiple options = fake with multiple checkboxes
-                    foreach (string allowedValue in configField.AllowValues)
+                    foreach (string value in configField.AllowValues)
                     {
                         configMenuIntegration.AddCheckbox(
-                            label: $"{configName}.{allowedValue}",
+                            label: $"{configName}.{value}",
                             description: null,
-                            get: config => configField.Value.Contains(allowedValue),
-                            set: (config, val) =>
+                            get: config => configField.Value.Contains(value),
+                            set: (config, selected) =>
                             {
-                                InvariantHashSet values = new InvariantHashSet(configField.Value);
-                                if (val && !values.Contains(allowedValue))
-                                    values.Add(allowedValue);
-                                if (!val && values.Contains(allowedValue))
-                                    values.Remove(allowedValue);
+                                // toggle value
+                                if (selected)
+                                    configField.Value.Add(value);
+                                else
+                                    configField.Value.Remove(value);
 
-                                if (values.Count == 0 && !configField.AllowBlank)
-                                    values = configField.DefaultValues;
+                                // set default if blank
+                                if (!configField.AllowBlank && !configField.Value.Any())
+                                    configField.Value = configField.DefaultValues;
 
-                                configField.Value = values;
+                                // update token
                                 resetToken();
                             }
                         );
@@ -557,11 +558,10 @@ namespace ContentPatcher
                     configMenuIntegration.AddDropdown(
                         label: configName,
                         description: null,
-                        get: config => configField.Value.Count > 0 ? configField.Value.First() : "",
-                        set: (config, val) =>
+                        get: config => configField.Value.FirstOrDefault() ?? "",
+                        set: (config, newValue) =>
                         {
-                            InvariantHashSet values = new InvariantHashSet(val);
-                            configField.Value = values;
+                            configField.Value = new InvariantHashSet(newValue);
                             resetToken();
                         },
                         choices.ToArray()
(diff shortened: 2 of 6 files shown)


ContentPatcher/ModEntry.cs
```diff
@@ -356,15 +357,43 @@ private void LoadContentPacks(IEnumerable<RawContentPack> contentPacks, Invarian
                         // load config.json
                         InvariantDictionary<ConfigField> config = configFileHandler.Read(current.ManagedPack, content.ConfigSchema, current.Content.Format);
                         configFileHandler.Save(current.ManagedPack, config, this.Helper);
+                        GenericModConfigMenuIntegration<InvariantDictionary<ConfigField>> configMenuIntegration = null;
                         if (config.Any())
+                        {
                             this.Monitor.VerboseLog($"   found config.json with {config.Count} fields...");
 
+                            configMenuIntegration = new GenericModConfigMenuIntegration<InvariantDictionary<ConfigField>>(
```
Pathoschild: I try to keep integration logic out of the main mod code. Can you move the new code out of `ModEntry` into a separate class similar to the existing `GenericModConfigMenuIntegrationForFastAnimations`/`…Tractor` classes?

ContentPatcher/ModEntry.cs
```diff
@@ -356,15 +357,43 @@ private void LoadContentPacks(IEnumerable<RawContentPack> contentPacks, Invarian
...
+                            if ( configMenuIntegration.IsLoaded )
+                                configMenuIntegration.RegisterConfig();
+                        }
+
                         // load config tokens
                         foreach (KeyValuePair<string, ConfigField> pair in config)
                         {
                             ConfigField field = pair.Value;
-                            IValueProvider valueProvider = new ImmutableValueProvider(pair.Key, field.Value, allowedValues: field.AllowValues, canHaveMultipleValues: field.AllowMultiple);
+                            ManualValueProvider valueProvider = new ManualValueProvider(pair.Key);
```
Pathoschild: This will break config tokens used in immutable fields like `Enabled`. We should probably keep these as `ImmutableValueProvider`, and remove/add the token when we change the config.

ContentPatcher/Framework/ModTokenContext.cs
```diff
@@ -146,7 +146,7 @@ public void UpdateContext(InvariantHashSet globalChangedTokens)
             // note: since token values are affected by the order they're defined, only updating tokens affected by globalChangedTokens is not trivial.
             foreach (ManagedManualToken managed in this.DynamicTokens.Values)
             {
-                managed.ValueProvider.SetValue(null);
+                managed.ValueProvider.SetValue((ITokenString)null);
```
Pathoschild: This conflicts with some CP 1.17 changes; can you rebase your changes onto the latest `develop`?

ContentPatcher/ModEntry.cs
```diff
@@ -463,5 +492,71 @@ select value.Trim()
...
+                if (configField.AllowMultiple)
+                {
+                    // Whitelist + multiple options = fake with multiple checkboxes
+                    foreach (string allowedValue in configField.AllowValues)
+                    {
+                        Func<InvariantDictionary<ConfigField>, bool> get = (config) => configField.Value.Contains(allowedValue);
+                        Action<InvariantDictionary<ConfigField>, bool> set = (config, val) =>
+                        {
+                            InvariantHashSet values = new InvariantHashSet(configField.Value);
+                            if (val && !values.Contains(allowedValue))
```
Pathoschild: * I suggest renaming `allowedValue` to `value`, and `val` to `selected` for clarity.
* Since we'll immediately set the `Value` field to the new set, we can just change it directly.
* We can simplify a bit, since `HashSet` will ignore duplicates and doesn't mind if you try removing a value that's not there. 

```cs
// old code
InvariantHashSet values = new InvariantHashSet(configField.Value);
if (val && !values.Contains(allowedValue))
    values.Add(allowedValue);
if (!val && values.Contains(allowedValue)) 
    values.Remove(allowedValue);
...
configField.Value = values;
```
```cs
// suggested code
if (selected)
   configField.Value.Add(value);
else
   configField.Value.Remove(value);
```

ContentPatcher/ModEntry.cs
```diff
@@ -463,5 +492,71 @@ select value.Trim()
...
+                                values.Add(allowedValue);
+                            if (!val && values.Contains(allowedValue)) 
+                                values.Remove(allowedValue);
+
+                            if (values.Count == 0 && !configField.AllowBlank)
+                                values = configField.DefaultValues;
```
Pathoschild: We should copy `DefaultValues` here to avoid passing a reference to the original field:
```cs
values = new InvariantHashSet(configField.DefaultValues);
```

ContentPatcher/ModEntry.cs
```diff
@@ -463,5 +492,71 @@ select value.Trim()
...
+                }
+                else
+                {
+                    // Whitelist + single value = drop down
+                    // Need an extra option when blank is allowed
+                    List<string> choices = new List<string>(configField.AllowValues);
+                    if (configField.AllowBlank)
+                        choices.Insert(0, "");
+
+                    Func<InvariantDictionary<ConfigField>, string> get = (config) => configField.Value.Count > 0 ? configField.Value.First() : "";
```
Pathoschild: This can be simplified:
```cs
// current code
configField.Value.Count > 0 ? configField.Value.First() : ""

// suggested
configField.Value.FirstOrDefault() ?? ""
```

ContentPatcher/ModEntry.cs
```diff
@@ -463,5 +492,71 @@ select value.Trim()
...
+                    configMenuIntegration.AddDropdown(configName, null, choices.ToArray(), get, set);
+                }
+            }
+            else
+            {
+                // No whitelist = text field
+                Func<InvariantDictionary<ConfigField>, string> get = (config) => string.Join( ", ", configField.Value.ToArray());
+                Action<InvariantDictionary<ConfigField>, string> set = (config, val) =>
+                {
+                    InvariantHashSet values = configField.AllowMultiple ? this.ParseCommaDelimitedField(val) : new InvariantHashSet(val);
```
Pathoschild: Allowing commas in a value may cause odd effects, since Content Patcher often assumes comma = separate value. Is it possible to pass a filter to GMCM so prohibited characters can't be typed instead?

Pathoschild (changes requested):
Thanks! I added some initial code review comments inline.

ContentPatcher/ModEntry.cs
spacechase0: Not at the moment. I could add a RegisterSimpleOptionFilter(IManifest manifest, string optionName, Func<string, string> filter). Would that work?

ContentPatcher/ModEntry.cs
Pathoschild: Done.

ContentPatcher/Framework/ModTokenContext.cs
Pathoschild: Done.

ContentPatcher/ModEntry.cs
Pathoschild: Done.

ContentPatcher/ModEntry.cs
Pathoschild: Done.

ContentPatcher/ModEntry.cs
Pathoschild: Done.

ContentPatcher/ModEntry.cs
Pathoschild: I changed the logic to only save the first value instead if multiple values aren't allowed, which is good enough for now and should be reasonably intuitive.

ContentPatcher/ModEntry.cs
Pathoschild: Done.

Pathoschild (approved):

Pathoschild:
Thanks! Refactored a bit and merged into `develop` for the upcoming Content Patcher 1.18.

PR #623, opened 2021-03-02
Update German translations for Data Layers

psychochicken80:
Translated tile grid


psychochicken80:
I think I messed up here... Sorry...

PR #637, opened 2021-04-26
Add shipping menu support to Fast Animations

### LINKED ISSUE #323: [Fast Animations] speed up end-of-day animations
Speed up the end-of-day animations (e.g. shipping screen boxes appearing, date change transition, etc).

[Suggested on Nexus](https://www.nexusmods.com/stardewvalley/mods/1089/?tab=posts&jump_to_comment=63095591).

adnan-shoukfeh:
[https://github.com/Pathoschild/StardewMods/issues/323](url)
Allows for speeding up end of day animation #323

### DIFF
--- a/FastAnimations/Framework/GenericModConfigMenuIntegrationForFastAnimations.cs
+++ b/FastAnimations/Framework/GenericModConfigMenuIntegrationForFastAnimations.cs
@@ -249,8 +249,8 @@ internal class GenericModConfigMenuIntegrationForFastAnimations : IGenericModCon
                 max: maxSpeed
             )
             .AddNumberField(
-                label: "Shipping menu transitions",
-                description: "How fast the shipping menu transitions between screens. Default 1x.",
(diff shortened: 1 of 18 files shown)


Pathoschild:
I made a few changes to fix the handler not applying.

The caveat is that we need a specialized event to run handlers on the shipping screen, and SMAPI will show a warning in the console:
> Bypassed safety checks
> --------------------------------------------------
> These mods bypass SMAPI's normal safety checks, so they're more likely to cause errors or save
> corruption. If your game has issues, try removing these first.
> 
> - Fast Animations

Before I merge this, I want to see if SMAPI can raise the normal events on the shipping screen (before the actual save process starts) in an upcoming version, so we can avoid the warning.

Pathoschild:
Thanks for the pull request! The upcoming SMAPI 4.2.0 now raises the normal update tick events on `ShippingMenu`, so this is now merged into `develop` for the upcoming Fast Animations 1.14.0.

PR #650, opened 2021-07-27
When clicking "add to existing stacks" do it for all storages

almightyju:
Not sure if harmony is the best way for this but it's certainly the easiest.

Whenever the button is clicked for a chest we simply call the same function for all other storage that use ItemGrabMenu (I don't have a save with all the various types so I can't 100% test this but even on a fail it should just revert to the default handling).


Pathoschild:
Thanks for the pull request! Unfortunately I'd prefer to avoid the Harmony dependency, and I think that feature is out of scope for Chests Anywhere itself. You could update [Convenient Chests](https://smapi.io/mods/#Convenient_Chests) though, or turn that feature into a new mod; then players could use both mods together.

PR #653, opened 2021-08-11
Suppress keybinds for use with binding to Controller Buttons

Deflaktor:
Added Suppress Keybind for:
- Lookup Anything Hotkeys
- Horse Flute Hotkey

This allows setting these Hotkeys to a DPad direction or the RightStick without making the character move or opening the chat window

### DIFF
--- a/HorseFluteAnywhere/ModEntry.cs
+++ b/HorseFluteAnywhere/ModEntry.cs
@@ -94,8 +94,13 @@ namespace Pathoschild.Stardew.HorseFluteAnywhere
         /// <param name="e">The event data.</param>
         private void OnButtonsChanged(object sender, ButtonsChangedEventArgs e)
         {
-            if (this.SummonKey.JustPressed() && this.CanPlayFlute(Game1.player))
-                this.HorseFlute.Value.performUseAction(Game1.currentLocation);
+            if (this.SummonKey.JustPressed() && Context.IsPlayerFree) {
+                this.Helper.Input.SuppressActiveKeybinds(this.SummonKey);
+                if (this.CanPlayFlute(Game1.player))
+                {
+                    this.HorseFlute.Value.performUseAction(Game1.currentLocation);
+                }
+            }
         }
 
         /// <summary>The event called after the location list changes.</summary>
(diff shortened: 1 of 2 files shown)


Pathoschild (approved):

Pathoschild:
Thanks! Merged into `develop` for the upcoming Horse Flute Anywhere 1.1.8 and Lookup Anything 1.35.2.

Deflaktor (commented):
Hm, if it only surpresses when it applies it does not cover my use case.
i have a xbox controller and would like to map the flute to dpaddown since I dont use the dpad at all. But if it only surpresses when it applies, then i walk down when pressing the dpaddown instead of getting the error message that I dont have the flute. If I override the dpaddown I dont want it to retain its old function at all.

Deflaktor:
So, could you revert 7761e62 ?

Pathoschild:
`CanPlayFlute` doesn't check whether the flute will work, only if you _could_ play it in the current context. So it won't suppress the keybind if you're not holding a flute at all (in balanced mode) or in a cutscene for example; but it _will_ suppress it if trying to play the flute shows an error (e.g. not enough space to summon a horse).

Deflaktor:
So I understand correctly, that if I don't have the flute in balanced mode, I can still use dpaddown to move downwards? I think it's weird that the button changes its behavior depending on whether I do own the flute or not.

Pathoschild:
Yep. Only suppressing input when the mod handles it is the usual convention in Stardew Valley modding, since there's many different contexts where input can be handled. If you want to always force-suppress a keybind, you could maybe create a simple mod which just has a list of keybinds in its `config.json` that it always suppresses.

Deflaktor:
I see. I guess in certain contexts it makes sense to follow that convention. 
I already tried making a tiny mod using a configurable list of keys to be surpressed. However, contrary to what the modding wiki says, it also prevents mods from handling the input. This is what brought me to make this PR in the first place.

Deflaktor:
Nvm, I was indeed able to make that simple mod for suppressing keys after taking a second look and soon after I found out about [EventPriority]
Turns out my mod was handling the suppress key stuff before the horse flute mod and as such it prevented it from firing. 

I now added a ButtonsChanged event with [EventPriority(EventPriority.Low - 10000)] so that it suppresses the key after all other mods may have handled it. This solves my use case.

Thanks for pointing me into the right direction!

PR #674, opened 2021-12-22
Fix transparency in sprites drawn in Lookup Anything menu

TehPers:
Sprites that are being drawn in the Lookup Anything menu have a weird alpha blending function being applied to them. Dimmed sprites (for example, in a recipe that isn't unlocked) clip through the background of the menu and partially show the world behind them:

image

This PR fixes the blend state that LU uses to draw the fields in its menus. It maintains the same color blending function (`srcColor * srcAlpha) + (dstColor * (1 - srcAlpha))`) but uses the destination alpha instead now to ensure that after the sprite is drawn, the alpha channel has the same value as the background it's being drawn on previously did.

image

### DIFF
--- a/LookupAnything/Components/LookupMenu.cs
+++ b/LookupAnything/Components/LookupMenu.cs
@@ -327,7 +327,17 @@ namespace Pathoschild.Stardew.LookupAnything.Components
                     {
                         // begin draw
                         device.ScissorRectangle = new Rectangle(x + gutter, y + gutter, (int)contentWidth, (int)contentHeight);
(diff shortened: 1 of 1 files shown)


Pathoschild (approved):

Pathoschild (approved):

TehPers:
Thanks for fixing the usage of `BlendState` Pathos! I missed that it was disposable, my bad.

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Lookup Anything 1.36.2.

PR #689, opened 2022-01-17
Update README.md

bajuba:
fixed a few grammar errors


Pathoschild:
Thanks for the pull request! The first one is a deliberate [Oxford comma](https://en.wikipedia.org/wiki/Serial_comma) though, and the second one is correct too since it's a conditional sentence.

PR #692, opened 2022-01-21
Feature/quick stack button

BananaFruit1492:
Implemented the following 2 features:

**1. QuickStackButton**
Inspired by the [Convenient Inventory Mod](https://www.nexusmods.com/stardewvalley/mods/10384), I implemented a quick stack button which tries to transfer all stackable items from the player's inventory to the available chests, if any of these chests contains items that are stackable with the items in the inventory.
This should save time pushing all stackable items to all chests.

For this feature I added the following config options controlling which chests are considered when determining the set of chests items can be moved to:
"_EnableQuickstackToAllAvailableChests_": true,
"_QuickStackOptions_": {
    "_ConsiderJunimoHuts_": false,
    "_ConsiderShippingBin_": false,
    "_ConsiderMiniShippingBins_": false,
    "_ConsiderAutoGrabber_": false
},

**2. Search functionality**
Upon hovering and scrolling over an item in the inventory, the next (scroll up) or last (scroll down) chest in the list of available chests containing at least one item with the same name as the hovered item is selected.
This should save time finding chests containing the item being hovered.


Pathoschild:
Thanks for all the work!

The chest search sounds interesting, and I think it could fit well with Chests Anywhere. I'd probably want the UI to work a bit differently so it's less hidden though. For example, maybe a UI similar to [Lookup Anything's search menu](https://user-images.githubusercontent.com/230581/150895850-5419a69e-8ea1-4299-8886-5526d2369928.png), or a filter button / search box which filters the dropdowns and navigation?

Unfortunately the quick stack feature is out of scope for Chests Anywhere, since the goal is really to simplify access. Ideally that would be handled by a separate mod like Convenient Inventory which focuses on doing that one thing well. (Sorry for all the work you already put into it!)

PR #698, opened 2022-01-25
Add Korean translations for Automate

wally232:
Korean translation file update.


Pathoschild:
Thanks for the help! You accidentally edited `default.json` instead, but I'll add the translations in the upcoming Automate 1.25.3.

PR #718, opened 2022-03-04
Add sections to Content Patcher's ConfigSchema

Shockah:
I wanted to keep it as backwards-compatible as possible, so I decided to just add a new `Section` field in ConfigField.

If a `config.section.<name>.name` translation is provided, it will be used for the visible name; otherwise `<name>` will be used.
If a `config.section.<name>.description` translation is provided, it will be passed as the tooltip for the section; otherwise the section will not have a tooltip. Allowing tooltips without i18n usage is not possible.

The GMCM integration will first add all config fields which don't have a section specified, then will go one-by-one through the sections, add a section title, then add that section's fields.

I leave updating the documentation to you, as you'll definitely do a much better job than me at that.


Pathoschild (approved):

Pathoschild:
Thanks for the pull request! I'll merge this into Content Patcher 1.26.0 as soon as development starts on that version.

Pathoschild:
Merged into `develop` for the upcoming Content Patcher 1.26.0. Thanks again!

PR #721, opened 2022-03-24
Correct config docs for section translations

ellipszist:


Pathoschild (approved):

Pathoschild:
Merged into `develop`, so the correction is now live. Thanks for the help!

PR #723, opened 2022-04-18
Allow fuzzier matching for types in `patch export filename type`

atravita-mods:
Currently, to patch export an asset that's not loaded, the user has to specify the fully qualified type name. For system types, this isn't too bad, but for images, it's... `"Microsoft.Xna.Framework.Graphics.Texture2D, MonoGame.Framework"`, which I'm pretty sure I've never gotten right on the first try. This PR does two things:

1. Allows users to type in `image` to refer to an image asset (case-insensitive).
2. Allows fuzzier matching when referring to type names. Just the full name (without the assembly) is sufficient, as long as there's not two types with the same full name.


Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Content Patcher 1.26.0.

PR #729, opened 2022-05-25
[Lookup Anything] Fix generic `HoveredItem` field support for custom sub-pages of `GameMenu`

KhloeLeclair:
The item lookup provider's menu handler currently detects if the active menu is `GameMenu` and gets the active page from it. Except for generic menu support.

The `HoveredItem` field for generic mod menu support is accessed on `menu` rather than `targetMenu`, thus not handling GameMenu children properly. This fixes that.


Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Lookup Anything 1.37.2.

PR #731, opened 2022-06-03
When loading patches from an IncludePatch, have reIndex off

SinZ163:
In my current modpack, it was previously taking 34 seconds (plus logging overhead) to do the initial UpdateTicked event.
With this change, it goes down to 5 seconds, and 3.6 when the logging overhead is removed.

This is partially untested, particularly when LoadPatches is invoked via ReloadCommand or by ApplyNewConfig

### DIFF
--- a/ContentPatcher/Framework/Patches/IncludePatch.cs
+++ b/ContentPatcher/Framework/Patches/IncludePatch.cs
@@ -141,7 +141,7 @@ namespace ContentPatcher.Framework.Patches
                         rawPatches: content.Changes,
                         rootIndexPath: this.IndexPath,
                         path: this.GetIncludedLogPath(this.FromAsset),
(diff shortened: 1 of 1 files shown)


Pathoschild:
Thanks for the pull request!

As we discussed on Discord, we'd need a few other changes since it depends when the include patch is being loaded. `reindex: false` is fine if it's happening as part of a parent patch load, but if it's part of a general context update (e.g. an `Include` patch with conditions) we'd need to reindex in that case.

Instead of tweaking the reindex logic though, I rewrote it in Content Patcher 1.26.6 (36ce44d76ebf68b1ee65693b119964c574279209 and 5d87a4f75e1b4fbdff43e22365a14b5a587ffd05) so the indexes are updated incrementally.

PR #741, opened 2022-06-23
Add RemoveDelimited text operation

Shockah:
This PR adds new Remove(First/Last/All)Occurrence(s) text operations. These can be used for example to remove a universal love (which would let you redefine it as another level of a gift).

I think I've done all the required things for this feature to be complete:
* Actual implementation of the feature
* Migration checking (won't allow using the operations unless the Format is 1.28.0)
* Validation (checking if `Delimiter` is provided, as it's required for these new text operations)
* Markdown documentation

### DIFF
--- a/ContentPatcher/ContentPatcher.csproj
+++ b/ContentPatcher/ContentPatcher.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <Version>1.27.1</Version>
+    <Version>1.28.0</Version>
     <RootNamespace>ContentPatcher</RootNamespace>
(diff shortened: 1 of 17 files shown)


Pathoschild:
Thanks! This seems fine overall at a quick glance. There's a few little things I'd tweak (e.g. add the missing method docs, and `Remove*` probably shouldn't convert null values to blank), but I can change those after I merge it.

I'm not planning to release Content Patcher 1.28 for a while though, since I'm mainly focused on stabilizing the recent changes and continuing with the performance improvements. So this might stay open for a bit, but it'll most likely be merged for 1.28!

Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Content Patcher 1.28.0.

PR #751, opened 2022-08-04
[Lookup Anything] Fix child birthday in future

iBug:
When using F1 on a child, its birthday is shown as a future date, and this shown date advances by 2 for every in-game day. This PR fixes that by removing one of the double negations (the other is `.AddDays(-daysOld)` in the next few lines).

The same issue also applies to the `stable` branch as the code is identical.


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Lookup Anything 1.37.4. Thanks for the help!

PR #755, opened 2022-08-25
Rewrite lexer to avoid regex

atravita-mods:
Hi Pathos!

This PR just moves the lexer off using regex to manually splitting the string using IndexOfAny. I haven't fully benchmarked it, but from watching Profiler on a decently sized mod group, the UpdateTicked just after GameLaunched goes from ~6.5 s to about 5.5s. So not something to write home about, but decently worthwhile, I think?


Pathoschild:
Thanks for the help! I'll put this in Content Patcher 1.29.0 since it's a bigger change, so I'll keep this open until then.

Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Content Patcher 1.28.2 or 1.29.0. Thanks again!

PR #782, opened 2023-01-08
Add map support to patch export

atravita-mods:
Hi Pathos!

This PR just adds maps to `patch export`. It'll also export the relevant tilesheets so the map can be opened fine in the `patch export` directory without having to fuss around with copying tilesheets.


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Content Patcher 1.29.0 or 2.0.0. Thanks for the help!

PR #790, opened 2023-02-16
Add optimizations to RangeValueProvider

SinZ163:
Adds the concept "IsDeterministicForInput" which is a softer version of !IsMutable which allows TokenString to keep it marked as immutable.

In addition, improves Condition.UpdateContext to only re-split the Values if its mutable (or empty to prevent it never being populated).

I'd love feedback on this if there was a better way of getting these savings.

Logs, all with extra profiler logging in place
|Log Name|Log URL|Notes|
|---|---|---|
|Before|[SMAPI.io](https://smapi.io/log/d82b60d4904b4d0da09acf1fd8416486?Levels=debug~critical&Mods=Profiler&Filter=handling+GameLoop.TimeChanged+%28)|N/A
|IsDeterministicForInput|[SMAPI.io](https://smapi.io/log/c80991a3577c413d861086176a864382?Levels=debug%7Ecritical&Mods=Profiler&Filter=handling+GameLoop.TimeChanged+%28)|Does not contain Condition changes
|Condition Changes|[SMAPI.io](https://smapi.io/log/4409f19734884fabb405ceb97d17ad44?Levels=debug%7Ecritical&Mods=Profiler&Filter=handling+GameLoop.TimeChanged+%28)|What is seen in this PR (minus profiler)

### DIFF
--- a/ContentPatcher/Framework/Conditions/TokenString.cs
+++ b/ContentPatcher/Framework/Conditions/TokenString.cs
@@ -110,7 +110,10 @@ namespace ContentPatcher.Framework.Conditions
                     tokensUsed ??= new();
                     tokensUsed.Add(token.Name);
 
-                    isMutable = isMutable || token.IsMutable;
+                    if (!token.RequiresInput || !token.IsDeterministicForInput)
+                    {
+                        isMutable = isMutable || token.IsMutable;
+                    }
                 }
                 else
                 {
--- a/ContentPatcher/Framework/Conditions/TokenString.cs
+++ b/ContentPatcher/Framework/Conditions/TokenString.cs
@@ -110,8 +110,11 @@ namespace ContentPatcher.Framework.Conditions
                     tokensUsed ??= new();
                     tokensUsed.Add(token.Name);
 
-                    if (!token.RequiresInput || !token.IsDeterministicForInput)
+                    if (!token.IsDeterministicForInput)
                     {
+                        // If every token is deterministic, then the resulting string must be immutable. If this token
+                        // is deterministic but it receives non-deterministic input, the string will be marked mutable
+                        // when we recursively check those tokens.
                         isMutable = isMutable || token.IsMutable;
                     }
                 }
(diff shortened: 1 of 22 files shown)


ContentPatcher/Framework/Conditions/TokenString.cs
```diff
@@ -110,7 +110,10 @@ public TokenString(IEnumerable<ILexToken> lexTokens, IContext context, LogPathBu
                     tokensUsed ??= new();
                     tokensUsed.Add(token.Name);
 
-                    isMutable = isMutable || token.IsMutable;
+                    if (!token.RequiresInput || !token.IsDeterministicForInput)
+                    {
+                        isMutable = isMutable || token.IsMutable;
+                    }
```
Pathoschild: Unfortunately we can't ignore token mutability just because it doesn't require input. For example, `{{Time}}` doesn't have input but still changes its values every seven seconds or so. (The token can also have optional input even if `RequiresInput` is false.)

I also suggest adding a comment to explain the optimization here.

```suggestion
                    if (!token.IsDeterministicForInput)
                    {
                        // If every token is deterministic, then the resulting string must be immutable. If this token
                        // is deterministic but it receives non-deterministic input, the string will be marked mutable
                        // when we recursively check those tokens.
                        isMutable = isMutable || token.IsMutable;
                    }
```

ContentPatcher/Framework/Tokens/ValueProviders/TranslationValueProvider.cs
```diff
@@ -25,7 +25,7 @@ internal class TranslationValueProvider : BaseValueProvider
         /// <summary>Construct an instance.</summary>
         /// <param name="translationHelper">Gets translations from the content pack's translation folder.</param>
         public TranslationValueProvider(ITranslationHelper translationHelper)
-            : base(ConditionType.I18n, mayReturnMultipleValuesForRoot: false)
+            : base(ConditionType.I18n, mayReturnMultipleValuesForRoot: false, isDeterministicForInput: true) // This is deterministic 99% of the time, but if language is changed, this would need to be invalidated
```
Pathoschild: Unfortunately this will introduce a bug where content packs don't update correctly when the player changes language on the title screen.
```suggestion
            : base(ConditionType.I18n, mayReturnMultipleValuesForRoot: false)
```

Pathoschild (changes requested):
Thanks for the PR! It looks good overall, there's just two bits we should review.

Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Content Patcher 1.30.0. Thanks for the help!

PR #796, opened 2023-04-17
Add migrations for some content assets changed in Stardew Valley 1.6

SinZ163:
Tentatively for Content Patcher 2.0 (unclear if 1.6 is getting that version bump or not).

Error messages still need improving to be more human friendly, and some edge cases still need handling, but it can read my 1.5.6 modlist with 3 exceptions (all for the same reason).

Log when it runs (Ignore the billion warnings from CP, they are likely due to a custom SMAPI build to ignore dependencies so I could run the migrator on as much as possible)
https://smapi.io/log/81ca6c3c7daf49749697ba0f18c5f70e
and the patch exported Data/Characters https://smapi.io/json/none/ba8015287b64466f9580fe117a507740

### DIFF
--- a/ContentPatcher/Framework/RawContentPack.cs
+++ b/ContentPatcher/Framework/RawContentPack.cs
@@ -20,7 +20,7 @@ namespace ContentPatcher.Framework
         private ContentConfig? ContentImpl;
 
         /// <summary>The backing field for <see cref="Migrator"/>.</summary>
-        private IMigration? MigratorImpl;
+        private IRuntimeMigration? MigratorImpl;
 
 
         /*********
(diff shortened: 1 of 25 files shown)


ContentPatcher/Framework/RawContentPack.cs
```diff
@@ -99,16 +99,6 @@ public bool TryReloadContent([NotNullWhen(false)] out string? error)
             IMigration migrator = new AggregateMigration(content.Format, this.GetMigrations(content));
             if (!migrator.TryMigrateMainContent(content, out error))
                 return false;
-            if (content.Changes.Any())
```
SinZ163: This isn't required as it runs in PatchLoader.LoadPatches, so this causes old migrations for content.json to run *after* newer migrations, as they are all now running twice.

SinZ163: Fixed upstream now, no longer an issue

Pathoschild (approved):

Pathoschild:
Thanks for the help!

I added a new 'runtime migrations' concept (which lets us support features like load patches, `TargetField`, text operations, tokens, etc), and converted these into runtime migrations.

Merged into `develop` for the upcoming Content Patcher 2.0.0.

PR #798, opened 2023-04-27
Add support for Fish in the alpha

SinZ163:
This doesn't quite give the new fish model justice, but it at least runs again and mostly has 1.5.6 parity restored


Pathoschild:
I unmarked as draft and merged it into `alpha`, since it's needed to use the new `DataLoader`. Thanks for the help!

PR #808, opened 2023-07-23
Make patch export better at guessing which type to export as

atravita-mods:
Hi Pathos!

This is an attempt to make patch export better guess at which type to use when it's given an asset path in the base game. It's mostly meant to handle the Maps/ case for now, but can be extended to handle other models (especially the Dictionary<string, ComplicatedModel> cases coming in 1.6)

It's mostly meant for base game things, although I guess mods can patch that method.

Sample log: https://smapi.io/log/acab765534654eec9f9e35d953549a3d?Levels=trace%7Edebug%7Einfo%7Ewarn%7Eerror%7Ealert%7Ecritical

### DIFF
--- a/ContentPatcher/Framework/Commands/Commands/ExportCommand.cs
+++ b/ContentPatcher/Framework/Commands/Commands/ExportCommand.cs
@@ -98,15 +98,45 @@ namespace ContentPatcher.Framework.Commands.Commands
             if (type is null)
                 return;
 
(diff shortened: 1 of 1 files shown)


Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Content Patcher 1.29.4.

Issue #194, opened 2018-02-28
[Tractor Mod] fuel mechanics


Pathoschild:
Consider adding optional fuel mechanics for players who want more challenge or immersion.

Requested a few times, most recently [2018-28](https://community.playstarbound.com/threads/better-tractor-mod.141385/#post-3257575).

Pathoschild:
Possibly add a fuel gauge (refilled with coal?) and water supply gauge (refilled with a watering can); when they drop to zero, the tractor stops moving and watering crops respectively.

compro01:
Suggestion : consider "fuelling" with battery packs as an option.

Pathoschild:
Proposed approach by DonUsernametaken on the Nexus page:

> Balanced mode that uses player energy whenever you press down the activation key. [...] Like drain 1 energy and then 1 more energy for every 2 seconds you hold it down, configurable in the config.

firereverie:
> Possibly add a fuel gauge (refilled with coal?) and water supply gauge (refilled with a watering can); when they drop to zero, the tractor stops moving and watering crops respectively.

I thought about trying to hook the tractor with PFM to do this, but I don't know what I'm doing yet, so I don't know how viable of an idea that is.

Would love to see
  config options for fuel/water bars or (I love the idea of the tractor emoting a frowny face) tractor status emotes.

  CP exposed options for fuel type, fuel efficiency & emote icons.

ZeroOne010101:
Hi!
I'm not a programmer by trade, but i tried my hand at adding fuel in the form of energy drain.

My first thought was to add a drain equivalent to the actions as if they had been done with the tools.
I didnt find a function to do that without copy-pasting the calcuations from the game, however, and i couldnt figure out how to implement it in a way that didnt trigger every update.
There'd need to be some way to get a list of eligible actions taken every tick i think.

I've now implemented a constant drain of x energy per second (one per default) with the corresponding entries in the mod menu by using a second delay, similar to the update one.
I'm fairly sure the button idea could be worked into that as an alternative mode.

Im unsure whether its a good idea to include however, as 1.6 is around the corner and I've read that the stamina attribute will be read-only.

Pathoschild:
@ZeroOne010101 Thanks for the pull request!

[...]

PR #812, opened 2023-07-30
[Tractor Mod] Adds constant energy drain.

ZeroOne010101:
Hello!
This might be a possible or partial solution to #194.

These proposed changes add a constant energy drain while the Tractor is mounted.
They also add menu bindings to toggle and modify said energy drain.

This uses the stamina property of the player, so the timing isn't ideal with the upcoming 1.6 changes.

My first thought was to add a drain equivalent to the actions as if they had been done with the tools.
I didnt find a function to do that without copy-pasting the calcuations from the game, however, and i couldnt figure out how to implement it in a way that didnt trigger every update.
There'd need to be some way to get a list of eligible actions taken every tick i think.

### DIFF
--- a/TractorMod/Framework/GenericModConfigMenuIntegrationForTractor.cs
+++ b/TractorMod/Framework/GenericModConfigMenuIntegrationForTractor.cs
@@ -436,6 +436,22 @@ namespace Pathoschild.Stardew.TractorMod.Framework
                     set: (config, value) => config.StandardAttachments.Slingshot.Enable = value
                 )
 
+                .AddSectionTitle(I18n.Config_Fuel)
+                .AddCheckbox(
+                    name: I18n.Config_FuelEnabled_Name,
+                    tooltip: I18n.Config_FuelEnabled_Tooltip,
+                    get: config => config.FuelEnabled,
+                    set: (config, value) => config.FuelEnabled = value
+                )
+                .AddNumberField(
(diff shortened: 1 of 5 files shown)


ZeroOne010101:
I am very sorry for not discussing those changes beforehand.
I mainly wanted the functionality for myself, since i am playing the game for the first time and am very close to buying the shed right now.
I think i may have jumped the gun there a little.

Ill leave this open for now, but please dont feel pressured to include it or anything.

Pathoschild:
Thanks for the pull request!

I'll close this per my comment in #194:
> I'm not sure I want to link the tractor fuel to the player's stamina though. While that does simplify the code, I think it would somewhat break immersion (since driving a tractor shouldn't be _that_ exhausting) and the ideal meta is to use the tractor in the morning (thus draining fuel) before spending the afternoon doing other stuff (thus draining stamina).
> 
> Your implementation is pretty similar to what I had in mind for the fuel mechanic though. It's just that instead of draining the player stamina, the mod would track a remaining-fuel value in the tractor's `modData` field.
>
> One small complication is that the upcoming Tractor Mod 4.17.0 (for Stardew Valley 1.6) adds tractor sound effects, so we'll need to adjust sounds for the fuel mechanic too (e.g. refueling, fuel runs out, riding tractor while out of fuel).

I might take another look at this when implementing the full fuel mechanic though. Thanks again!

PR #817, opened 2023-08-21
Update Chinese translations for Automate

Annijang:


Pathoschild:
Thanks for the pull request!

Unfortunately those translations were also updated in #811. I'll merge that one since it was submitted first, but feel free to submit a new pull request with any improvements you want to suggest to those. Thanks again!

PR #824, opened 2023-10-28
Simplify Lookup Anything level progression bars

TehCupcakes:
First of all, thanks for this awesome set of mods! LookupAnything is one of the most useful mods I've found, and ever since I got it years ago I can't imagine playing Stardew without it. 😄 

However, just as long as I have been using the mod, this issue has been nagging me, so I decided to do something about it. On the player info window, it draws experience bars for your progress in each skill. The problem is that it calculates the progress for _each_ bar relative to your current level's exp value! This looks really awkward with multiple bars partially filled. Comparison of before and after:

LookupAnythingExpBarFix


Pathoschild (approved):

Pathoschild:
Thanks for the help! Technically it was deliberate, but it often caused confusion so I'm fine with the change. Merged into `develop` for the upcoming Lookup Anything 1.40.3.

PR #840, opened 2023-12-22
Fix 'Error Item' preserve recipes, fix all objects except coffee beans being treated as seeds

SinZ163:
This fixes all objects (except Coffee Seeds) being treated as a Seed, along with fixing Error Preserves due to multiple (O) getting appended and becoming error items.

Before:
image
image

After:
image
image

### DIFF
--- a/LookupAnything/DataParser.cs
+++ b/LookupAnything/DataParser.cs
@@ -446,25 +446,22 @@ namespace Pathoschild.Stardew.LookupAnything
             {
                 if (inputID != null)
                 {
-                    switch (ItemRegistry.GetDataOrErrorItem(outputID).ItemId)
+                    var objectDefinition = ItemRegistry.GetObjectTypeDefinition();
+                    switch (ItemRegistry.QualifyItemId(outputID))
                     {
-                        case "342":
-                            obj.preserve.Value = SObject.PreserveType.Pickle;
+                        case "(O)342":
+                            obj = objectDefinition.CreateFlavoredPickle(ItemRegistry.Create(inputID) as SObject);
                             break;
-                        case "344":
-                            obj.preserve.Value = SObject.PreserveType.Jelly;
+                        case "(O)344":
+                            obj = objectDefinition.CreateFlavoredJelly(ItemRegistry.Create(inputID) as SObject);
                             break;
-                        case "348":
-                            obj.preserve.Value = SObject.PreserveType.Wine;
+                        case "(O)348":
+                            obj = objectDefinition.CreateFlavoredWine(ItemRegistry.Create(inputID) as SObject);
                             break;
-                        case "350":
-                            obj.preserve.Value = SObject.PreserveType.Juice;
+                        case "(O)350":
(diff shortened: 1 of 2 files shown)


LookupAnything/DataParser.cs
```diff
@@ -445,30 +445,26 @@ private Item CreateRecipeItem(string? inputID, string outputID, MachineRecipeOut
             {
                 if (inputID != null)
                 {
-                    switch (outputID)
+                    switch (ItemRegistry.GetDataOrErrorItem(outputID).ItemId)
                     {
-                        case "(O)342":
                         case "342":
```
Pathoschild: We should use the qualified item ID instead, to avoid conflicts with other `SObject`-type item IDs (e.g. furniture).

In that case we can also change `ItemRegistry.GetDataOrErrorItem(outputID).ItemId` to `ItemRegistry.QualifyItemId(outputId)`.

LookupAnything/DataParser.cs
```diff
@@ -445,30 +445,26 @@ private Item CreateRecipeItem(string? inputID, string outputID, MachineRecipeOut
...
                             obj.preserve.Value = SObject.PreserveType.Pickle;
-                            obj.preservedParentSheetIndex.Value = inputID;
```
Pathoschild: We can create flavored items using the game's new methods for it now, instead of doing it manually. For example:
```cs
return ItemRegistry
    .GetObjectTypeDefinition()
    .CreateFlavoredPickle(ItemRegistry.Create(inputId));
```

LookupAnything/Framework/Lookups/Items/ItemSubject.cs
```diff
@@ -467,10 +467,14 @@ private string GetTypeValue(Item item)
 
             try
             {
-                Crop crop = new(seed.ItemId, 0, 0, location ?? Game1.getFarm());
```
Pathoschild, line 470: Similarly I suggest changing the condition above to avoid item ID collisions:
```cs
// current
if (seed is not SObject obj || obj.bigCraftable.Value)

// suggested
if (seed.HasTypeId(ItemRegistry.type_object))
```

Pathoschild (changes requested):
Thanks for the help! Just a few comments before it's merged.

LookupAnything/DataParser.cs
SinZ163: Done

LookupAnything/Framework/Lookups/Items/ItemSubject.cs
SinZ163: Done (inverted HasTypeId to preserve the short circuit behavior)

LookupAnything/DataParser.cs
SinZ163: Done, though doesn't short circuit return to continue allow MachineRecipeOutputData to continue overriding if it exists

Pathoschild (approved):

Pathoschild:

PR #841, opened 2023-12-30
Some attachments can have a larger radius than others

NermNermNerm:
Some tools, particularly watering and scythe, get used a lot in farms that are built around using the tractor.  It's really nice to have those tools have a bigger impact area than tools like the hoe, where you'd rather have more precision.

This adds a new setting "ExtendedDistance" to the Scythe, Waterer, Seeder and Fertilizer tools.  It's just a boolean setting which allows for one more than the default radius.  It's a simple boolean setting to make it simpler for the user.  Any area of affect greater than the default 3x3/5x5 throws out all notions of game balance anyway, so the main 'Distance' setting will be good enough for people who want to go bigger.


TractorMod/Framework/Config/ScytheConfig.cs
```diff
@@ -29,5 +29,8 @@ internal class ScytheConfig
 
         /// <summary>Whether to clear weeds.</summary>
         public bool ClearWeeds { get; set; } = true;
+
+        /// <inheritdoc/>
```
NermNermNerm: Whoops, I used this out of habit, when I shouldn't have since it's not used elsewhere.  I can revert to a copy/paste of the interface if you want.

(I'd recommend you look into using this everywhere to cut down on duplicate comments.)

TractorMod/Framework/Config/GenericAttachmentConfig.cs
```diff
@@ -1,6 +1,6 @@
 namespace Pathoschild.Stardew.TractorMod.Framework.Config
 {
-    /// <summary>Configuration for an attachment which can only be enabled or disabled.</summary>
+    /// <summary>Configuration for an attachment which has no extra modalities - just on or off.</summary>
```
NermNermNerm: I can revert this change if you want.  Not sure why I thought it was better...

TractorMod/Framework/Config/StandardAttachmentConfig.cs
```diff
@@ -7,7 +7,7 @@ internal class StandardAttachmentsConfig
         public AxeConfig Axe { get; set; } = new();
 
         /// <summary>Configuration for the fertilizer attachment.</summary>
-        public GenericAttachmentConfig Fertilizer { get; set; } = new();
+        public ExtendedDistanceConfig Fertilizer { get; set; } = new() { IncreaseDistance = false };
```
NermNermNerm, line 10: For fertilizer and seeding, I felt like it fits the "non-destructive" idea, but as for playing this way all the time, I'm not so sure.  It's a rarely done activity, so personally, I care more about precision than speed.  So I made it something that folks could turn on if they felt differently about it, but left if off by default.

TractorMod/Framework/ExtendedDistanceAttachment.cs
```diff
@@ -0,0 +1,23 @@
+using Pathoschild.Stardew.TractorMod.Framework.Config;
+using StardewModdingAPI;
+using StardewValley;
+
+namespace Pathoschild.Stardew.TractorMod.Framework
+{
+    internal abstract class ExtendedDistanceAttachment<TConfig> : BaseAttachment
```
NermNermNerm, line 9: missing comments

TractorMod/Framework/Config/ScytheConfig.cs
NermNermNerm: I went ahead and reverted it.  Still, strongly suggest using <inheritDoc>

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/xmldoc/recommended-tags#inheritdoc

(Though I concede it'd be better to do it at a stroke rather than piecemeal.)

TractorMod/Framework/Config/GenericAttachmentConfig.cs
NermNermNerm: reverted

TractorMod/Framework/Config/ScytheConfig.cs
Pathoschild: The code does use `inheritdoc` in a lot of places already, so it'd be fine to use it here since you added the interface. I'll do that before merging though.

Pathoschild:
We discussed this on Discord, but I'll just summarize here for the record.

I'm planning to change the approach so that, instead of each attachment having an `IsDistanceExtended` bool option to use a second distance value, players will be able to override the distance per-attachment. In Generic Mod Config Menu, it'd appear as a distance slider where the default value is "use default distance".

I'm mostly working on compatibility updates until the dust settles on the Stardew Valley 1.6 update, so I'll mark this as a draft until we can make those changes.

Pathoschild:
I'd want to reimplement this per my earlier comment, but it's been in draft for over a year and I haven't seen players asking for a feature like this. So unfortunately I don't have any current plans to work on it. Thanks for the PR though, I do appreciate the contribution even if it ended up not being merged!

PR #842, opened 2024-01-05
Fix fertilizer translations and missing crafting recipes

SinZ163:
Previously, Fertilizer translations were failing due to double qualifying when trying to get the display name.
(Note: Crop and Error Wine are both fixed in #840)
image
Now, it correctly shows the value.
image


Pathoschild (approved):

Pathoschild:

PR #843, opened 2024-01-05
Fix crafting/cooking recipes missing when queried by ingredient

SinZ163:
This later causes mismatches in RecipeIngredientModel.Matches(Item item) as it does Qualified Id string comparisons.

(Note crop info on an iridium bar is fixed by #840)
Previous:
image
Now:
image


SinZ163:
Change to explicitly Qualify when building the RecipeIngredientModel blows up on categories, switching to a different approach

Pathoschild (approved):

Pathoschild:

PR #863, opened 2024-03-26
Fix runtime migrations adding spaces to context tags

SinZ163:
This fixes issues like the mayo machine not making mayo anymore if contexttags ever gets migrated.
As currently all but the first tag gets replaced with a version starting with a space, which doesn't get matched with machine logic anymore.

Prior to the change:
image


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Content Patcher 2.0.2. Thanks for the help!

PR #873, opened 2024-03-29
Use CanApplyFertilizer to check if fertilizer can be placed on that tile instead of HasFertilizer

foxwhite25:
CanApplyFertilizer is the method that the base game hand planting and green box rendering use, it can more reliably indicate whether fertilizer can be placed on that tile or not, and thus improve mod compatibility ~(with my mod)~. This also fixes a bug/feature, you can place fertilizer after a crop have grown with tractor, which is not a vanilla behaviour, player can install additional mod to alter vanilla behaviour if they want to


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Tractor Mod 4.17.3. Thanks for the help!

PR #877, opened 2024-03-30
Automate now harvests moss from trees

skoliosaurus:
Hi there, 
while playing stardew with the automate mod I noticed that the moss from trees wasn't being harvested.
Since I was curious about creating mods for stardew valley I decided to see if I could add the feature myself, and it seems to work.

I've never created stardew mods before, and it's been at least 5 years since I've last used C#, but decided to create a PR anyway, since it might still be useful.

thanks for the great mods and the many hours you've put into creating them


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Automate 2.0.4. Thanks for the help!

Issue #882, opened 2024-04-03
[Lookup Anything] Missing info on Largemouth Bass


gr3ger:
**Describe the bug**
Largemouth Bass (and maybe more pages?) does not have any data showing season/time/location of availability like other fish do.
If I find the problem I'll make a PR, but I'm just creating an issue just in case since I don't know when I'll be able to sit down and take a look at it.

**To Reproduce**
1. Shift+F1 and search for largemouth bass
2. everything besides how to get the fish is shown

**Log file**
https://smapi.io/log/c43f52aaebcf49309552b5eebfd49c27

**Screenshots**
![image](https://github.com/Pathoschild/StardewMods/assets/456624/4dd1788d-4250-463b-9fa7-ea1940310404)

EDIT:
Seems like this issue happens with all fish that don't care about seasons. I've added further info down below in the comments.

gr3ger:
Updated the log with one with a leaner modlist. Issue persists regardless though.

gr3ger:
![image](https://github.com/Pathoschild/StardewMods/assets/456624/34b0b809-f88b-466b-979f-c740471b0cce)
It does have tags for lake, all seasons, and no weather favorability at least. It just doesn't show in-game.

gr3ger:
It seems like the Largemouth bass falls under a third state that the current if/elseif doesn't capture here:
https://github.com/Pathoschild/StardewMods/blob/bbc4b7a81ef6d11faa7085767ba79943ed077396/LookupAnything/DataParser.cs#L165-L183

I've just done some quick probing.
But the current code detects that it should spawn in Mountain and Backwoods (?) during the first phase, but fails to actually add any locations to the list due to the season/condition checks.

gr3ger:
I think I may have narrowed it down.
Bullhead and Carp exhibit the same issues.

So I think all fish with no season constraints have empty locations in the mod.
Adding a third "else" to handle such cases could work if there's a way to validate that the locations are actually correct.

gr3ger:
If I understood everything correctly, then maybe this is correct after all?
If there are no conditions attached to the fish, it means there's not outstanding circumstances for it to spawn.

If there's only one season it is available in, it would fall into the `Season` attribute of SpawnFishData.
If there were more seasons, or other more complex conditions it would fall under the `Condition` attribute.
But since it's available in all seasons and aren't dependent on anything else, Condition *should* be null, and Season *should* be empty for these fishes.

gr3ger:
[...]

PR #883, opened 2024-04-03
[Lookup Anything] Fix fishing locations

gr3ger:
I opened an issue earlier today #882 about fish that are available all the year around seems to have issues showing their location. 
This fixes the issue, however I am unsure if this is the root cause or even the correct way to do it, so please point me in the right direction in that case since it has been a hot minute since I was in the stardew modding scene. 

Three fish I found that exhibited issues:
Largemouth Bass
Bullhead
Carp

### DIFF
--- a/LookupAnything/DataParser.cs
+++ b/LookupAnything/DataParser.cs
@@ -137,6 +137,7 @@ namespace Pathoschild.Stardew.LookupAnything
         /// <remarks>Derived from <see cref="GameLocation.getFish"/>.</remarks>
         public FishSpawnData? GetFishSpawnRules(string fishID, Metadata metadata)
         {
+            Console.WriteLine("Getting spawn rules for fish: " + fishID);
             // get raw fish data
             string[] fishFields;
             {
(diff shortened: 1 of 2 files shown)


gr3ger:
I just noticed a discord message you wrote about it. 
If the data for spawn rules got borked in 1.6, point me in the right direction as to where it happens and I can take a look there instead since this might not be a proper fix.

gr3ger:
If I understood everything correctly, then maybe this is correct after all?
If there are no conditions attached to the fish, it means there's not outstanding circumstances for it to spawn.

If there's only one season it is available in, it would fall into the `Season` attribute of `SpawnFishData`.
If there were more seasons, or other more complex conditions it would fall under the `Condition` attribute.
But since it's available in all seasons and aren't dependent on anything else, `Condition` *should* be null, and `Season` *should* be empty for these fishes.

Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Lookup Anything 1.41.4. Thanks for the help!

PR #889, opened 2024-04-04
Content Patcher managed token string API

spacechase0:
Will add some basic docs once I test this


ContentPatcher/Framework/Api/ApiManagedValue.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            this.ManagedValues = new PerScreen<IManagedValue>(parse);
+        }
+
+        /// <inheritdoc />
+        public IEnumerable<int> UpdateContext()
+        {
+            return new HashSet<int>(
+                this.ManagedValues
+                    .GetActiveValues()
```
Pathoschild: This is _technically_ a bug since it'll update all the values based on the current player's context. For example, `{{PlayerName}}` will return the same value for all players. We have the same issue with `ApiManagedCondition` though, so we can just keep it as-is and fix both of them separately.

ContentPatcher/IManagedValue.cs
```diff
@@ -0,0 +1,33 @@
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace ContentPatcher
+{
+    /// <summary>A set of parsed values linked to the Content Patcher context. These values are <strong>per-screen</strong>, so the result depends on the screen that's active when calling the members.</summary>
```
Pathoschild: I suggest changing "_a set of parsed values linked to..._" to something like "_a parsed string which may contain Content Patcher tokens linked to..._" for clarity.

spacechase0:
Test code I used:
```csharp
        private IManagedValue test;
        private void GameLoop_UpdateTicked(object sender, StardewModdingAPI.Events.UpdateTickedEventArgs e)
        {
            var cp = Helper.ModRegistry.GetApi<IContentPatcherApi>("Pathoschild.ContentPatcher");
            if (cp.IsConditionsApiReady && test == null)
            {
                test = cp.ParseValues(ModManifest, "{{Season}} is my favorite season", new SemanticVersion("2.0.0"));
            }

            if (test != null)
            {
                test.UpdateContext();
                Log.Info("Meow! " + (test.IsValid ? ( test.IsReady ? test.Value : "not ready") : test.ValidationError));
            }
        }
```

ContentPatcher/Framework/Api/ApiManagedValueForSingleScreen.cs
```diff
@@ -0,0 +1,100 @@
...
+        /*********
+        ** Fields
+        *********/
+        private readonly IManagedTokenString TokenString;
+
+        /// <summary>The context with which to update conditions.</summary>
+        private readonly IContext Context;
+
+        /// <summary>A contextual manager for the underlying conditions.</summary>
+        private readonly AggregateContextual Contextuals;
```
Pathoschild: `AggregateContextual` collects multiple `IContextual` values into a single group which can be updated and checked together. But in this case we only have one value, so I suggest just using the relevant methods on `this.TokenString` directly instead.

ContentPatcher/Framework/ContentPatcherAPI.cs
```diff
@@ -97,6 +103,21 @@ public IManagedConditions ParseConditions(IManifest manifest, IDictionary<string
...
+        public IManagedValue ParseValues(IManifest manifest, string rawValue, ISemanticVersion formatVersion, string[]? assumeModIds = null)
+        {
+            // validate lifecycle
+            if (!this.IsConditionsApiReady)
+                throw new InvalidOperationException($"'{manifest.Name}' accessed Content Patcher's conditions API before it was ready to use. (For mod authors: see the documentation on {nameof(IContentPatcherAPI)}.{nameof(IContentPatcherAPI.IsConditionsApiReady)} for details.)");
+
+            // validate dependency on Content Patcher
+            if (!manifest.HasDependency(this.ContentPatcherID, out ISemanticVersion? minVersion, canBeOptional: false))
+                throw new InvalidOperationException($"'{manifest.Name}' must list Content Patcher as a required dependency in its manifest.json to access the values API.");
+            if (minVersion == null || minVersion.IsOlderThan("1.22.0")) // TODO Change me
```
Pathoschild: This would be `2.1.0`, assuming it's added in the next Content Patcher major update.

ContentPatcher/IManagedValue.cs
Pathoschild: We should update/add the code docs to refer to values instead of conditions (including in the other files too).

Pathoschild (commented):
Thanks for the PR!

I'll probably merge this into the upcoming Content Patcher 2.1.0, but I'm mainly working on compatibility updates for Stardew Valley 1.6 for now. Here's a few comments based on a quick review in the meantime; feel free to ignore these if you prefer and I'll handle them after I merge.

ContentPatcher/Framework/ContentPatcherAPI.cs
spacechase0: I left that for you to change since I wasn't sure when you'd want to release it, but I can go ahead and change it to that.

spacechase0:
Pushed the docs and changes. There was one thing I wasn't sure about, I'll comment on the corresponding line.

ContentPatcher/Framework/Api/ApiManagedValueForSingleScreen.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        /*********
+        ** Private methods
+        *********/
+        /// <summary>Get whether the conditions need to be updated for the current context.</summary>
+        private bool ShouldUpdate()
+        {
+            // update once if immutable
+            if (!this.TokenString.IsMutable)
+                return false;
```
spacechase0: Is this correct now without AggregateContextuals, or do I need to make a corresponding way to track WasEverUpdated?

spacechase0:
Uh oh I missed something and this doesn't compile, hold on.

spacechase0:
Okay, fixed that line.

ContentPatcher/Framework/Api/ApiManagedValueForSingleScreen.cs
Pathoschild: We don't need a `WasEverUpdated` in this case since the token string already manages that; I tweaked the code to reuse `TokenString.ShouldUpdate()` instead.

Pathoschild:
Merged into `develop` for the upcoming Content Patcher 2.1.0. Thanks for the help!

PR #901, opened 2024-04-10
Update item retrieval method in BushMachine

LeFauxMatt:
This alone does not accomplish Custom Bush integration because it doesn't account for the stack or quality, but at least produces the correct (modded) item without affecting the original vanilla logic.


LeFauxMatt:
This alone does not accomplish Custom Bush integration because it doesn't account for the stack or quality, but at least produces the correct (modded) item.

Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Automate 2.0.7. Thanks for the help!

PR #902, opened 2024-04-10
Add integration with Custom Bush mod

LeFauxMatt:
This allows Lookup Anything to draw from the correct texture for custom bush, and loads the Name/Description from the Custom Bush api. To get the next harvest will essentially be copying code from FruitTreeSubject since CustomBush uses a data model that mirrors FruitTree.


LeFauxMatt:
This allows Lookup Anything to draw from the correct texture for custom bush, and loads the Name/Description from the Custom Bush api. To get the next harvest will essentially be copying code from FruitTreeSubject since CustomBush uses a data model that mirrors FruitTree.

Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Lookup Anything 1.42.0. Thanks for the help!

PR #918, opened 2024-04-23
[Lookup Anything] Fix flavored item names in preserve jar & keg lookups, and incorrect caviar recipes in roe lookups

### LINKED ISSUE #913: Cannot lookup "Preserves Jar" item after 1.41.6 update.
Cannot lookup "Preserves Jar" item after 1.41.6 update.

**Log file**
https://smapi.io/log/e06118f55c2d416ba4e3202cf955788e

**Screenshots**

b3nk3lly:
Fixes https://github.com/Pathoschild/StardewMods/issues/913, which happens on lookups on Preserve Jars due to a `NullReferenceException`. The exception is thrown when roe is created as a recipe item in `ItemRecipeField.TryCreateItemEntry`. I don't know why but updating `preserve.Value` on the roe breaks its display name property, which causes the exception.

I added a check to `TryCreateItemEntry` so we no longer update `preserve.Value` for roe. This fixes the lookup on Preserve Jars, but it reveals another problem with the names of the generic jelly and pickles:

preserves-jar-bugged-recipes

I found a similar error on the lookup for Kegs, this time with the names for the generic wine and juice:

keg-bugged-recipes

These are happening in the `CreateRecipeItem` method because when we do a lookup on a Keg or Preserve Jar, the machine becomes the ingredient. We are trying to make a flavored jelly / pickle out of a Preserve Jar and a flavored wine / juice out of a Keg.

To fix the issue I added checks to `CreateRecipeItem`  so that we don't try to create a flavored item out of Kegs or Preserve Jars. I also added logic for when to create flavored aged roe, because I was seeing a bug where looking up any roe would give the caviar recipe.

Here are the lookups for Preserve Jars and Kegs after these changes:

keg-fixed-recipes

preserves-jar-fixed-recipes

### DIFF
--- a/LookupAnything/DataParser.cs
+++ b/LookupAnything/DataParser.cs
@@ -451,11 +451,13 @@ namespace Pathoschild.Stardew.LookupAnything
             Item output;
             switch (outputId)
             {
-                case "(O)342":
+                // if ingredient is Preserves Jar, we want generic pickles instead of flavored
+                case "(O)342" when ingredient.ItemId != "15":
                     output = ItemRegistry.GetObjectTypeDefinition().CreateFlavoredPickle(ingredient as SObject);
                     break;
 
-                case "(O)344":
+                // if ingredient is Preserves Jar, we want generic jelly instead of flavored
+                case "(O)344" when ingredient.ItemId != "15":
                     output = ItemRegistry.GetObjectTypeDefinition().CreateFlavoredJelly(ingredient as SObject);
                     break;
 
--- a/LookupAnything/DataParser.cs
+++ b/LookupAnything/DataParser.cs
@@ -448,25 +448,34 @@ namespace Pathoschild.Stardew.LookupAnything
         {
             outputId = ItemRegistry.QualifyItemId(outputId);
 
+            bool isIngredientKeg = ingredient.ItemId == "12";
+            bool isIngredientPreservesJar = ingredient.ItemId == "15";
+            bool isIngredientRoe = ingredient.ItemId == "812";
+
             Item output;
             switch (outputId)
             {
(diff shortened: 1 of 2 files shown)


LookupAnything/DataParser.cs
```diff
@@ -448,36 +448,47 @@ private Item CreateRecipeItem(Item ingredient, string outputId, MachineRecipeOut
...
-            {
-                obj.preservedParentSheetIndex.Value = outputData.PreservedParentSheetIndex ?? obj.preservedParentSheetIndex.Value;
-                obj.preserve.Value = outputData.PreserveType ?? obj.preserve.Value;
-            }
-
+                {
+                    obj.preservedParentSheetIndex.Value = outputData.PreservedParentSheetIndex ?? obj.preservedParentSheetIndex.Value;
+                    obj.preserve.Value = outputData.PreserveType ?? obj.preserve.Value;
+                }
+
```
kevinmarrec: Not sure about the added indentations, could be reverted.

kevinmarrec (changes requested):

LookupAnything/DataParser.cs
b3nk3lly: True my bad, I took em out, thanks !

Pathoschild (approved):

Pathoschild:
I made two changes:
- I removed the `NullReferenceException` workaround, since that was a game bug fixed in Stardew Valley 1.6.6.
- I tweaked the logic to avoid passing the machine as an input, which avoids needing to hardcode machine IDs in `CreateRecipeItem`.

Otherwise the PR looks good to me!

Pathoschild:
Merged into `develop` for the upcoming Lookup Anything 1.42.1. Thanks for the help!

PR #929, opened 2024-05-06
[LookupAnything] Add support for machine recipes from Data/Machines

zombifier:
LOOKUP

What's done:

- Lookup machine recipes that requires input and produces output, without taking conditions into account. When looking up an input item, will match all recipes that use it by `RequiredItemId`, `RequiredTags`, or `AdditionalConsumedItems`. Outputs are matched by getting all outputs from an output rule.
- If a recipe has conditions attached (either through trigger or output), add a "(conditional)" indicator text
- Integration with extra ingredients added by [ExtraMachineConfig](https://www.nexusmods.com/stardewvalley/mods/22256?tab=description) in #934
- PFM support in #933 (though I haven't tested this too closely)

Todo:

- Format code
- Testing

What's left (that won't be done in this PR):

- Better way of displaying tags
- Actually show conditions, `UseFirstValidOutput`
- Support trigger rules that aren't `ItemPlacedInMachine`
- Support `DROP_IN`


scriptsforweirdos:
Will this make it work with PFM machines as well? Because it currently does not.

zombifier:
Only CP for now. I can look into PFM next.

scriptsforweirdos:
That would be great! I'm working on the Raffadax update and while some machines are using Content Patcher, the majority are still in PFM.

Digus:
@zombifier , if you need anything for PFM api that is currently not working to support LA, just let me know.

zombifier:
@Pathoschild @Digus I've spun out the PFM changes into #933. This PR is pretty much done; any future changes and features listed above can be done in future commits/PRs.

Let me know if you want the Extra Machine Config integration spun out as well. I think it's pretty self-contained and easy to review.

Digus:
@zombifier , I think the hasCondition would be nice to add for PFM rules as well since PFM has it's own share of conditions. I probably would need to change the API to add that flag as well. LA would need to add those to ProducerFrameworkRecipe class and propagate it to RecipeModel.

Pathoschild:
@zombifier Thanks! I merged the PFM changes into `develop` for the upcoming Lookup Anything 1.43.0. And yep, splitting the remainder into two PRs would be ideal (one for reading the vanilla machine data, and one for the Extra Machine Config support).

zombifier:
@Pathoschild split EMC integration into #934. Marked that as draft for now since it's dependent on this one.

zombifier:
> @zombifier , I think the hasCondition would be nice to add for PFM rules as well since PFM has it's own share of conditions. I probably would need to change the API to add that flag as well. LA would need to add those to ProducerFrameworkRecipe class and propagate it to RecipeModel.

That would be ideal, though condition support isn't quite finalized yet. Perhaps your API can return the entire condition string in some form so it can be displayed in the lookup form

Digus:
>  That would be ideal, though condition support isn't quite finalized yet. Perhaps your API can return the entire condition string in some form so it can be displayed in the lookup form

PFM has no string condition, but a group of properties that combined and based on order factors can give the condition to rules. I could write them in a string so people can read those and try to interpreter when the output might be possible. Some are straightforward, like required input quality or stack, but other like RequiredMail, RequiredEvent, RequiredInputParentIdentifier (ex. requiring Orange based product instead of any), can be difficult to represent besides just writing. 

Output Probability can be really misleading, since PFM only make one roll when producing output, and check those in order adding up the results, so if 3 outputs is are available, one with .7 probability, one with 0.6 and other with 0.3. The later has no chance to happen, since the first one will have 70% and the second one will have a 30% chance even though their probability value is .6. Another thing that can happen is if 2 are available and one have 0.3 and the other 0.5, the first one will have 50% chance since renaming probability will go to the default output (the first one in the list of available outputs)
Another behavior is that if there are item with no probability set, they share the remaining chance. So if one output has .7 probability another .2 and two more have no probability, they will have .05 each.
Since output availability is calculated based on others requirements. Its impossible to know the real chance of each output before hand.
So I would just not show probability, and just required conditions. At first just written, but some could be visually represented, like quality and stack, but they would require these values to be passed separated so LA don't need to parse a string condition.

Pathoschild:
This works great for simple machine rules like the furnace:
furnace recipes

The current limitations I see are...
1. Context tags are shown as-is, which will be a bit confusing for players. We could build a list of items matching the context tags, though it might be a bit expensive to do that for every recipe (particularly with mods that heavily use context tags). It would also be tricky to support dynamic context tags (e.g. item quality).

   previous | experimental
   -------- | ------------
   bone mine recipes (before) | bone mill recipes (after)

2. [Game state queries](https://stardewvalleywiki.com/Modding:Game_state_queries) are shown as "conditional" with no info. We could maybe create a simple parser to convert some common queries into human-readable text, and fallback to showing the condition text as-is.

   previous | experimental
   -------- | ------------
   keg recipes (before) | keg recipes (after)

3. [Item queries](https://stardewvalleywiki.com/Modding:Item_queries) are represented by the first item they produced when we opened the lookup. That's fine for the current vanilla machines, which use simple queries which always produce the same item. It won't handle more dynamic queries like `RANDOM_ITEM` which mods likely use though.

   This one is pretty tricky, and I'm not sure if we can solve this one. We can maybe just show a disclaimer that some recipes may not be shown if we detect a dynamic or unknown query.

4. Min/max stack [quantity modifiers](https://stardewvalleywiki.com/Modding:Migrate_to_Stardew_Valley_1.6#Quantity_modifiers) aren't handled.

zombifier:
1. I'm inclined towards building a utility class/function for returning human readable version of category tags, the same way category numbers are being handled (e.g. `category_fruit,quality_iridium,!id_o_613` -> Fruit AND Iridium Quality AND NOT Apple). It won't be able to handle every category tags in existence (and any unsupported tags can be returned unchanged), but a set of vanilla tags can be supported and mod authors can maybe add their own.
2. Same with above, but this will be much, much more effort. Perhaps we can just stick the condition's full string somewhere in the window if it can't be parsed.
3. The code calls `ItemQueryResolver.TryResolve`, which should return every item resolvable by the query? I haven't tested a machine with that rule though, and you're right it would be very tricky if that's the case.

Pathoschild:
I think the main limitations are addressed now, though we can certainly polish it in later versions.

1. We now show translated human-readable representations for common context tags (and others will be shown as-is):
   mayonnaise machine lookup
2. Game state queries are now shown as-is (and we can add human-readable representations in future versions):
   keg lookup
3. I think the current approach for item queries is good enough for now. It covers all vanilla machines, and I'm not sure many custom machines will use non-deterministic item queries (and most custom ones will probably be day-stable like `DISH_OF_THE_DAY`).
4. We don't handle some advanced fields like stack/quality modifiers, but we can extend support in future versions.

Pathoschild (approved):

Pathoschild:

PR #932, opened 2024-05-09
Fix JSON syntax in Content Patcher EditData docs

AlanDavison:
Fairly self-explanatory -- we had someone confused in #making-mods.

Though this probably ought to be an issue of its own, but is there a reason `js` is being used instead of `json` for styling the code blocks?


Pathoschild (approved):

Pathoschild:
Merged into `develop`. Thanks for the help!

The docs use `js` syntax highlighting to allow non-strict JSON like comments:

`js` | `json`
---- | ------
js format example | json format example

PR #934, opened 2024-05-11
[LookupAnything] Add support for extra fuels added by ExtraMachineConfig

zombifier:
[Extra Machine Config](https://www.nexusmods.com/stardewvalley/mods/22256) allows machine recipes to define extra inputs aside from the main input and the machine's global fuel.

This mod's API has 2 functions:
- `GetExtraRequirements`: returns a map of fuel qualified ID to quantity for the specified output item rule.
- `GetExtraTagRequirements`: same as above, but the key is a context tag list instead.

To test, you can install [Visit Mount Vapius](https://www.nexusmods.com/stardewvalley/mods/9600). With EMC installed some recipes are modified to have different or multiple additional inputs:

- Jewelry Machine: Flower Globe use Glass Shard instead of Gold Bar like the rest of the recipes in the Jewelry Machine
- Preserves Jar: Tapenade needs Anchovy as additional input
- Loom: Night Silk needs bat wings
- Loom: Night Silk can also be made using silk flowers, bat wings and wool

### DIFF
--- a/Common/Common.projitems
+++ b/Common/Common.projitems
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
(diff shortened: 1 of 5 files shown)


Pathoschild (approved):

Pathoschild:

PR #939, opened 2024-05-12
[Lookup Anything] In Progression Mode, hide recipes until they are discovered

### LINKED ISSUE #655: [Lookup Anything] Do not show recipes until acquiring them
So yeah, would be nice if ProgressionMode also hid recipes which havent been acquired yet.

b3nk3lly:
Closes #655, which suggests that recipes stay hidden until they are discovered if the player is in Progression Mode.

This change hides recipes in two sections of the lookup page: the "Needed for" and "Recipes" sections.
- In the "Needed for" section, recipe names are hidden in the requirements for Craft Master and Gourmet Chef. The text now says "cook/make [known recipes], X unknown recipes".
- In the "Recipes" section, the number of unknown recipes is displayed at the end of the list of known recipes. This count can appear for cooking, crafting, and tailoring recipes. It's a bit hard to tell in the screenshots, but the new text is gray instead of black, like the "X unrevealed villagers" text.

I also changed the text for the new collapsible fields on the lookup menu. If recipes are hidden due to progression mode, the count for "show X results" has been adjusted so it no longer counts the hidden recipes. So in the screenshot below for wood, it says "show 23 results" instead of the usual "show 52 results".

This change adds 1 new string that will need to be translated. Maybe we should consider adding a second string for when there is only a single hidden recipe, so we don't see "1 unknown recipes".

Screenshots:
leek-recipes-hidden
fibre-recipes-hidden
wood-recipes-hidden


Pathoschild (approved):

Pathoschild:

PR #942, opened 2024-05-17
Add support for building conversion recipes from Data/Buildings

zombifier:
MILLS

Screenshot uses vanilla and modded recipes from Cornucopia and Visit Mt Vapius.

Known issue: There are no input icons since apparently building rules only accept tags, not any specific item IDs. We'd also need to build a reliable method of mapping item ID context tags back to the respective items.


Pathoschild:
With my changes on the branch, context tags which uniquely identify an item are now converted to item IDs and show the icon:
>

Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Lookup Anything 1.43.0. Thanks again for the help!

PR #943, opened 2024-05-17
Update Japanese translations for Lookup Anything

mitekano23:
Untranslated sections have been updated. Please check it. If there are no problems, I would be happy to have them added along with the other elements when they are updated.


Pathoschild (approved):

Pathoschild:

PR #946, opened 2024-05-19
[Fast Animations] More Fast Animations

weizinai:
Speed up the chest-open animation.
Speed up the item-holdup animation.


weizinai:
The temporary sprite issue is not solved.

Pathoschild:
Hi! What do you mean by the temporary sprite issue?

weizinai:
I solved this problem based on HarvestHandler. At first, I only sped up the player's animation, which caused there to be temporary 
sprite of items stuck on the map.

weizinai:
But now there's another problem, the animation for holding up the item only lasts 3000 milliseconds, but pauses the player for 4000 milliseconds, which results in 1000 milliseconds before the player can move after the animation ends.

Pathoschild:
I suggest submitting a separate PR for each feature. That way we can release the ones that are ready, even if one of them is still in progress.

weizinai:
I have submitted a separate PR for each feature. I apologize, this is my first PR submission.

PR #960, opened 2024-05-21
[Fast Animations] Add Hold Up Item Handler

weizinai:
Speed up item-holdup animation.


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Fast Animations 1.12.0. Thanks again for the help!

PR #962, opened 2024-05-21
[Fast Animations] Add Use Totem Handler

weizinai:
Speed up use-totem animation.


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Fast Animations 1.12.0. Thanks again for the help!

PR #964, opened 2024-05-21
Fix items not being replaced when referenced by their old Qualified Ids

Digus:
The code was using their new Qualified Item Id to look for items in the game, instead of their old Qualified Item Id.


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Content Patcher 2.1.0. Thanks for the help!

PR #979, opened 2024-05-23
[LookupAnything] Fix machine lookup issues for mods that use both item IDs and context tags

zombifier:
The issue reported on the Nexus page (unrelated recipes showing up in item lookups) is caused by mods that add recipes that use both item IDs and context tags. The code currently assumes that only one will be used, and thus match every item that matches either one or the other, not both.

Additionally, change the lookup display field to also append context tags to the item ID if applicable.

To test, install the [Copper Still](https://www.nexusmods.com/stardewvalley/mods/15333) mod. Before this PR a couple of its recipes will show up in every item lookup because it defines a recipe with both `"RequiredItemId": "(O)348"` (Wine) and `"RequiredTags": [ "!preserve_sheet_index_90" ]` (not Cactus Fruit flavored), and the latter is matching with every item in the game. After this fix they should no longer show up, and will also display properly when the Still is looked up.

### DIFF
--- a/LookupAnything/Framework/Fields/ItemRecipesField.cs
+++ b/LookupAnything/Framework/Fields/ItemRecipesField.cs
@@ -440,9 +440,12 @@ namespace Pathoschild.Stardew.LookupAnything.Framework.Fields
 
                 if (input is not null)
                 {
+                    string name = input.DisplayName ?? input.ItemId;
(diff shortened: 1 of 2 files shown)


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Lookup Anything 1.43.2. Thanks again for the help!

PR #984, opened 2024-05-28
Update Chinese translations for Lookup Anything

weizinai:


LookupAnything/i18n/zh.json
```diff
@@ -130,24 +130,23 @@
...
-    "condition.item-edibility.range": "{{item}} has edibility of {{min}} to {{max}}",
+    "condition.item-edibility.edible": "{{item}}是可食用的",
+    "condition.item-edibility.min": "{{item}}的可食用性大于等于{{min}}",
+    "condition.item-edibility.range": "{{item}}的可食用性为{{min}}到{{max}}",
 
     "context-tag.bone": "骨头（任意）",
     "context-tag.egg": "鸡蛋（任意）",
     "context-tag.large-egg": "大鸡蛋（任意）",
-    "context-tag.preserved-item": "preserved {{name}}", // TODO
+    "context-tag.preserved-item": "{{name}}",
```
Pathoschild: Hi! This text is shown when a recipe needs a preserved item. For example, Blueberry Wine is 'preserved' blueberries. It will probably be confusing to show Blueberry in that case, since you can't use blueberries in the recipe directly.

Pathoschild:
Merged into `develop` for the upcoming Lookup Anything 1.44.1. Thanks again for the help!

PR #1006, opened 2024-06-22
[ChestsAnywhere] Add enricher to ChestFactory

Mushymato:
Technically only checks for object is sprinkler and object's held object's held object is a chest, never explicitly checks item id (because I desired compatibility with my mod).


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Chests Anywhere 1.24.0. Thanks for the help!

PR #1011, opened 2024-06-30
[LookupAnything] Don't show recipes for nonexistent machines

zombifier:
Some mods (ie. H&W Baking) has config options that toggles machines on and off while still leaving their recipe data in Data/Machines intact. Those recipes are, by proxy, disabled because there are no machines to process them, but they still show up in ingredient lookup regardless. This change stops those recipes from showing up.

(modders on Discord suggest still showing these if show debug is on. I'm not sure how I would implement that, so leaving it up to you whether you think the idea is useful).

### DIFF
--- a/LookupAnything/DataParser.cs
+++ b/LookupAnything/DataParser.cs
@@ -407,10 +407,7 @@ namespace Pathoschild.Stardew.LookupAnything
             {
                 string qualifiedMachineId = entryKey; // avoid referencing loop variable in closure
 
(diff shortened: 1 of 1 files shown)


Pathoschild (approved):

Pathoschild:
I think hiding them entirely is fine. Merged into `develop` for the upcoming Lookup Anything 1.45.2. Thanks for the help!

PR #1016, opened 2024-07-21
[ChestsAnywhere] Feature: Auto item sorting from anywhere

PedroLucasMiguel:
This is a new **optional** feature that allows the player to automatically sort all items from their inventory into chests from anywhere.

### How it works
Basically, we look into the player inventory and for each item, we try to find a **chest** that have a similar item stored. For now, I think that this feature only make sense to store items into **chests**, but this could be easily configurable.

After looking for a chest, we check the item stacks in order to try to store the item, the conditions to store an item are the following:

- If a chest have an item with the same name and quality, proceeds with the checking stage;
- If the chest item stack is full, search for other stack;
- If the chest item stack is half-full, store X items until the stack is full and search for other stack;
- If the stack can accept all the player items, stores than and stop checking for other stacks;

By the end of the process, we show a simple message to the player using `Game1.addHUDMessage` to any item was sorted or not.

### How it would be shipped
This feature could be quite "op", so it's shipped as disabled by default.

The player can also change the range of the item sorter, as well as the key to sort the items if they so desire.

### Regarding translations
As I only speak Portuguese and English, I only did the translations for those languages, so any help with the other languages would be great!

### Final words
That's my first time trying to add a feature to a mod, so any feedback is welcome :P

### DIFF
--- a/ChestsAnywhere/Framework/GenericModConfigMenuIntegrationForChestsAnywhere.cs
+++ b/ChestsAnywhere/Framework/GenericModConfigMenuIntegrationForChestsAnywhere.cs
@@ -77,6 +77,20 @@ namespace Pathoschild.Stardew.ChestsAnywhere.Framework
                     allowedValues: Enum.GetNames<ChestRange>(),
                     formatAllowedValue: name => I18n.GetByKey($"config.range.{name}")
                 )
+                .AddCheckbox(
+                    name: I18n.Config_AutoSorter_Name,
+                    tooltip: I18n.Config_AutoSorter_Desc,
+                    get: config => config.EnableAutoSort,
+                    set: (config, value) => config.EnableAutoSort = value
+                )
+                .AddDropdown(
+                    name: I18n.Config_AutoSorterRange_Name,
+                    tooltip: I18n.Config_AutoSorterRange_Desc,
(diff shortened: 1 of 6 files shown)


Pathoschild:
Thanks so much for the help!

Unfortunately I think this is out of scope for Chests Anywhere itself, which focuses on categorizing and accessing your chests. I prefer smaller mods that focus on doing their core feature well, so players can combine multiple small mods to get the specific set of features they want.

In this case, players would probably add a mod like [Quick Stash](https://www.nexusmods.com/stardewvalley/mods/20864) or [Stash Items](https://www.nexusmods.com/stardewvalley/mods/9624) to do this, or possibly an omnibus mod like [Better Chests](https://www.nexusmods.com/stardewvalley/mods/9791) (which doesn't follow the small-mod philosophy).

You're also welcome to convert this PR into a new stash mod which players could install alongside Chests Anywhere. I think the existing mods all focus on nearby chests, so this could be an added option for players. If you're interested, feel free to ping me on [the Stardew Valley Discord](https://smapi.io/community#Discord) and we could discuss making Chests Anywhere's `ChestFactory` available as an API to other mods, to simplify creating companion mods like that.

Sorry, and thanks again for the PR!

PR #1018, opened 2024-07-25
Add the ability to 'patch reload' a specific json in your content pack.

spacechase0:
This would mean you don't have to reload your entire content patch when only wanting to test changes from one file. (Larger content packs take a long time to reload, so this can be helpful for skipping most of that waiting time.)


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Content Patcher 2.4.0. Thanks for the help!

PR #1023, opened 2024-08-12
[LookupAnything] Hide recipes that involve invalid items

Mushymato:
Hide recipes that either require invalid items or produce an invalid output.

Shown again if "Show Datamine Fields" is enabled, maybe should be separate option?

Edit: now uses new config field "Show Invalid Recipes"


Pathoschild:
Yeah, I suggest adding a separate option for this. Ideally the 'show data mining fields' option should only do that.

Pathoschild (approved):

Pathoschild:

PR #1027, opened 2024-08-23
[LookupAnything] Don't sort machine recipe ingredients lists and keep the trigger item as the first item.

zombifier:
Machine recipes have a primary 'trigger' item and secondary 'fuel' items. We're already putting the trigger item first while building the machine rules; for clarity's sake also keep it as the first item in the lookup model, before the extra fuels.
Before:
2024-08-23T17:40:03,884328249-04:00
After:
2024-08-23T17:41:54,717276229-04:00


Pathoschild (approved):

Pathoschild:

PR #1028, opened 2024-08-23
[LookupAnything] Add integration with Extra Machine Config's multiple output items feature

zombifier:
[API Reference](https://github.com/zombifier/My_Stardew_Mods/blob/master/ExtraMachineConfig/Api/IExtraMachineConfigApi.cs)

For now basically add new recipes with the same ingredient list but new output items taken from EMC's API. It does mean the lookup can't distinguish between "the recipe can produce one of these items" and "the recipe will produce all of these items", but I don't think it's important to distinguish (and any solutions I can think of to distinguish them leads to combinatorics explosions if there are multiple possible outputs for each 'slots'). Plus, this is also how things currently work with building conversion recipes (e.g. Mill).

### DIFF
--- a/Common/Integrations/ExtraMachineConfig/IExtraMachineConfigApi.cs
+++ b/Common/Integrations/ExtraMachineConfig/IExtraMachineConfigApi.cs
@@ -16,5 +16,6 @@ public interface IExtraMachineConfigApi
 
     /// <summary>Retrieves the extra output items produced by this recipe.</summary>
     /// <param name="outputData">The output rule to check.</param>
+    /// <param name="machine">The machine data which contains the <paramref name="outputData"/>.</param>
     IList<MachineItemOutput> GetExtraOutputs(MachineItemOutput outputData, MachineData? machine);
 }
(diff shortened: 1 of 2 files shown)


zombifier:
Seems like this will get spammy with the 'global byproduct' feature that adds a byproduct to every machine recipes. I'll need to rethink how the UI should work (maybe multiple items on one line), but a workaround for now is probably to exclude those, and only add the per-recipe byproduce.

zombifier:
To recap, here are the solutions I'm thinking for a recipe that takes A and returns B as a primary output, and C and D as secondary:

1. (Existing) Show:
    - A => B
    - A => C
    - A => D
    Pros: simplest, most straightforward
    Cons: Doesn't communicate that C and D are byproduce from A to B, might bloat the screen for byproduce that applies to every recipe
2. Show:
    - A => B only
    - Looking up C and D will show A => B
    Pros: Least clutter
    Cons: May not be clear
3. Show:
    - A => B + C + D
    Pros: Clearest, best looking solution
    Cons: Extremely complicated, may lead to explosion of recipe counts if there are multiple items in the C and D slot

I'm still most happy with 1, but I'm interested to hear your thoughts.

zombifier:
Merged with latest develop, should work now (see screenshot for example from the new [Wildflour Atelier Goods](https://www.nexusmods.com/stardewvalley/mods/27049), which use this feature extensively. It can also be downloaded to test out).

2024-11-23T21:49:00,578091032-05:00

The recipe takes 2 milk and outputs all 4 of the items listed.

Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Lookup Anything 1.50.0.

PR #1031, opened 2024-09-10
[Lookup Anything] Fix counts for collapsible fields when some results are grouped

### LINKED ISSUE #1025: [Lookup Anything] Only count revealed elements when deciding whether to collapse a list
**Recommendation:**
List collapsibility limits should only count revealed elements.  At most, the "show ### results" should only contribute a small amount toward the collapsibility limit.

For example, when viewing a villager's gift tastes with the "Show unknown gift tastes" option unchecked, the limit for collapsing large fields should only respect known gift tastes.

**Rationale:**
Collapsing list items is invaluable for preventing UI clutter by enabling the user to "opt in" to the info dump they want to see.  This doesn't really apply when the uncollapsed view is still within readable limits.  It is not terribly helpful from a usability standpoint on Year 1 Spring 7 to see a see "Likes gifts:  show 648 results" on a villager's information screen only to click "show 648 results" and see "Pizza, 647 unrevealed items".

**Reproduction Steps**
Exact steps which reproduce the bug, if possible. For example:
1. Open Lookup Anything's mod configuration screen using Generic Mod Configuration Manager.
2. Uncheck "Show unowned gifts".
3. Check "Collapse large fields".
4. Set "NPC lookup: gift tastes" to 10.
5. Start a new save.
6. Wait until 12:00 pm.
7. Purchase a Salad from Gus.
8. Talk to Pierre.
9. Give Pierre the Salad.
10. Bring up the Lookup Anything lookup window for Pierre.

**Expected Behavior**
"Likes gifts" shows "Salad, # unrevealed items", where # is the total number of gifts Pierre likes minus one.

**Actual Behavior**
"Likes gifts" shows a link with the text "show # results", where # is the total number of gifts Pierre likes.  Clicking the link replaces it with "Salad, # unrevealed items", where # is the total number of gifts Pierre likes minus one.

**Screenshots**
image
image

b3nk3lly:
When deciding whether to collapse a field, instead of using the total number of items / recipes in the field, we now use the length of the field's `Value`. This way, grouped results (e.g., "647 unrevealed items", which might appear when using progression mode) are counted as one result. This closes https://github.com/Pathoschild/StardewMods/issues/1025.

A special case is when displaying recipes grouped by type. In this case, the length of `Value` is always 0, so we instead need to manually sum up all the displayed recipes in each group.

Example:
Collapsed fields, progression mode on (and collapse limit of 1, for the sake of demonstration):
image

Expanded fields, progression mode on:
image

Collapsed fields, progression mode off:
image

### DIFF
--- a/LookupAnything/Framework/Fields/CharacterGiftTastesField.cs
+++ b/LookupAnything/Framework/Fields/CharacterGiftTastesField.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Xna.Framework;
@@ -11,6 +12,13 @@ namespace Pathoschild.Stardew.LookupAnything.Framework.Fields
     internal class CharacterGiftTastesField : GenericField
     {
         /*********
+        ** Accessors
+        *********/
+        /// <summary>The total number of items shown (including the sum of grouped entries like "11 unrevealed tastes").</summary>
+        public int TotalItems { get; }
+
+
+        /*********
         ** Public methods
         *********/
         /// <summary>Construct an instance.</summary>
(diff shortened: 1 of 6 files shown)


Pathoschild (approved):

Pathoschild:

I made one change: after deciding whether to collapse the field per your PR, the label still shows the actual total number of items. For example, "_click to show 15 results_"  → "_1, 2, 3, 4, 5, 6, 7, 8, 9, 10, and 5 unrevealed gift tastes_". That better reflects the intended meaning of the message (e.g. "this field covers 15 gift tastes" rather than "this field covers 11 comma-delimited values"), but we could change it in later versions if that causes confusion.

PR #1036, opened 2024-09-22
[Lookup Anything] Add schedule to NPC lookup

### LINKED ISSUE #4: [Lookup Anything] show NPC schedule
When looking up an NPC, show their schedule for today. It looks like we can get today's schedule from `Dictionary<int, SchedulePathDescription> NPC::getSchedule(int dayOfMonth)`. Although it takes a day of month, it's only accurate for the current day since it checks the current conditions (e.g. rain).

[Discussed in the forums](http://community.playstarbound.com/threads/smapi-lookup-anything.122929/page-4#post-3036993).

b3nk3lly:
Added a field for schedules in the NPC lookup. Closes https://github.com/Pathoschild/StardewMods/issues/4.

Schedule items display the time they occur and the NPC's destination location. The schedule item that is currently occurring is written in green, upcoming items are in black, and past items are in gray. For schedules that contain only a single item, the string "All day" is shown instead of the time.

Some locations do not have built-in display names, like `LeoTreeHouse` and `HarveyRoom`. To fix this, I made `DataParser`'s `GetLocationDisplayName` method public and added resource strings for locations without built-in display names.

Screenshots:
image
image
image

### DIFF
--- a/LookupAnything/DataParser.cs
+++ b/LookupAnything/DataParser.cs
@@ -324,6 +324,38 @@ internal class DataParser
         return this.GetLocationDisplayName(fishSpawnData.LocationId, locationData, fishSpawnData.Area);
     }
 
+    /// <summary>Get the translated display name for a location and optional fish area.</summary>
+    /// <param name="id">The location's internal name.</param>
+    /// <param name="data">The location data, if available.</param>
+    /// <param name="fishAreaId">The fish area ID within the location, if applicable.</param>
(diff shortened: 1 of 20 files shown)


Pathoschild (approved):

Pathoschild:
Merged into `develop` (with a few changes) for the upcoming Lookup Anything 1.47.0. Thanks for all the work!

PR #1037, opened 2024-09-24
[Content Patcher] Clarify the documentation for editing lists with EditData

dmorsecode:
Editing a list of models with Content Patcher is a common source of confusion that I see due to the difference between the original asset having no keys (as it is, of course, a list) and the items added/edited within the Entries field needing keys regardless for Content Patcher to utilize. This PR adds on to the documentation for the EditData action to clearly show how editing a list is possible by way of example, and also makes it clearer that the ability to edit a list is separate from, and not only consisting of, the process for moving entries within a list.


Pathoschild:
Merged into `develop` for the upcoming Content Patcher 2.4.0. Thanks for the help!

PR #1039, opened 2024-09-25
[Content Patcher] Add ReplaceDelimited text operation

dmorsecode:
This PR adds a new text operation, ReplaceDelimited, that searches delimited strings to find values similar to RemoveDelimited except instead of removing the found values it will replace them with a user-provided value. This can be helpful for replacing specific words or values in the middle of a delimited string where the index of said value is still important and the other text operations won't suffice. The ReplaceMode field is also accepted here just as it is in RemoveDelimited.

It does accept null replacement values, in which case it will replace the found search value with an empty string and be functionally similar to RemoveDelimited, but enforcing non-null replacement values instead would be a simple change to make if accepting null values is not desired.


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Content Patcher 2.4.0. Thanks again for the help!

PR #1041, opened 2024-09-26
[LookupAnything] Fish spawn rule field should use qualified item ID

Mushymato:
Use qualified id to obtain fish spawn info, so that a big craftable like chest isn't showing tuna schedule.


Pathoschild (approved):

Pathoschild:
Thanks again for the help! Merged into `develop` for the upcoming Lookup Anything 1.46.0.

PR #1054, opened 2024-11-05
[LookupAnything] Add a continue after detecting complex output method

Mushymato:
The base ItemQueryResolver.TryResolve overload seems to be using Game1.log.Warn instead of Helpers.ErrorResult (which respects the logError delegate passed into TryResolve), leading to some warnings even when logError is null.

LookupAnything currently attempts to resolve item query for case where OutputMethod is set. Skip this as vanilla game does not try to resolve item query in this case, only apply item fields.

### DIFF
--- a/LookupAnything/DataParser.cs
+++ b/LookupAnything/DataParser.cs
@@ -497,8 +497,10 @@ internal class DataParser
                             continue;
 
                         // track whether some recipes are too complex to fully display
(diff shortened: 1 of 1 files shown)


Pathoschild:
Hi! It's still useful to track the recipe even if it can't be fully displayed (e.g. to know which recipes require the item), so I fixed it a different way in Lookup Anything 1.46.2. Thanks for the help though!

PR #1063, opened 2024-12-01
[Automate] Remove TapperMachine

zombifier:
`DataBasedObjectMachine` now does what `TapperMachine` does, making the latter redundant ([code reference](https://github.com/Pathoschild/StardewMods/blob/develop/Automate/Framework/Machines/DataBasedObjectMachine.cs#L130)). Remove the class and just have `DataBasedObjectMachine` handle the tapper logic.

This change is motivated by Machine Terrain Framework's custom tapper with rules in `Data/Machines`, which were having issues with Automate because it was expecting logic from `DataBasedObjectMachine` to fulfill (namely, repopulating the machine with `OutputCollected`), but was instead getting `TapperMachine`.

NOTE: I've kept the actual type for now because there are mods that are patching `TapperMachine` code (like Forage Fantasy). Your call as to whether to keep it around to not break mods, albeit making them non-functional, or remove the class to create visible errors.


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Automate 2.3.2. Thanks for the help!

PR #1070, opened 2025-01-10
Update Ukrainian translations for Chests Anywhere

EnderFS:
Translated the text that remained untranslated.
Made several corrections in the text that was already translated:
- Replaced small letters with capital letters in the words that were the first in the sentence;
- Corrected several words that incorrectly described the essence of the word in the context in which they were used;
- Text that was translated without taking into account the context.


Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Chests Anywhere 1.25.3.

PR #1075, opened 2025-01-14
Add integration with Iconic Framework

LeFauxMatt:
I plan to submit PRs for each of the respective mod integrations that exists today from Iconic Framework to the mods in this repo, but I wanted to submit just one for now to get your thoughts on any changes you would suggest so that I can remain consistent for the remaining PRs.


LookupAnything/ModEntry.cs
```diff
@@ -122,6 +123,19 @@ private void OnGameLaunched(object? sender, GameLaunchedEventArgs e)
...
+        IconicFrameworkIntegration iconicFramework = new(this.Helper.ModRegistry, this.Monitor);
+        if (iconicFramework.IsLoaded)
+        {
+            iconicFramework.AddToolbarIcon(
+                "LooseSprites/Cursors",
+                new Rectangle(330, 357, 7, 13),
+                I18n.Icon_ToggleSearch_Name,
+                I18n.Icon_ToggleSearch_Desc,
+                this.TryToggleSearch
+            );
```
Pathoschild: For Lookup Anything, maybe we should add a second icon to lookup something in front of the player? That could be useful for mobile players so they don't need to use a virtual keyboard.

LeFauxMatt:
I added secondary action support to the shortcut api method so that now the primary action is to lookup the grab tile, and the secondary action toggles search.

https://github.com/user-attachments/assets/b83371f2-fbf7-43ed-94a8-bb6aff1caf07

Pathoschild:
If the player enables tile lookups, that would prevent the search UI from ever being shown. I don't suppose Iconic Framework supports something like sub-actions (so the player could choose from 'lookup' or 'search')?

Something like this quick mockup:
sub-action mockup

LeFauxMatt:
Iconic Framework actually uses another mod Radial menu for this feature so that would be something it would need to add. Although, I don't quite understand why the search UI would be inaccessible to the player. The framework provides a primary and secondary action depending on which controller button they use to select the option. The secondary action is tied to Lookup Anything's `ModEntry.TryToggleSearch` which seems like it should only ever display the search UI regardless of Tile Lookup.

Pathoschild:
Nope, what you have is fine per our discussion on Discord. I initially thought it was falling back to the search UI automatically, not using a different input for each action.

Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Chests Anywhere 1.26.0, Data Layers 1.19.0, Debug Mode 1.15.0, Horse Flute Anywhere 1.3.0, Lookup Anything 1.48.0, and Tractor Mod 4.22.0.

PR #1081, opened 2025-02-11
[Automate] Refactor to support global inventory chests

LeFauxMatt:
This is an early draft of the changes to implement Global Inventory support for Automate.

### DIFF
--- a/Automate/Framework/Commands/Summary/GroupContainerStats.cs
+++ b/Automate/Framework/Commands/Summary/GroupContainerStats.cs
@@ -28,7 +28,7 @@ internal class GroupContainerStats
     public int TotalSlots { get; }
 
     /// <summary>Whether the container is a Junimo chest.</summary>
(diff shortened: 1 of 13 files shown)


LeFauxMatt:
@Pathoschild this not 100% working yet, but I wanted to submit an early draft to get your thoughts on the overall implementation so that I can make adjustments before I get too deep into it.

LeFauxMatt:
@Pathoschild as a proof of concept, this now works in the few basic tests I have done.

I setup a Chest with a GlobalInventoryId in one location, and setup another chest with the same GlobalInventoryId in a different location. I then connected a regular chest next to one of them with resources, and setup a machine next to the other one. The resources began to process as expected.

There are still a lot of changes needed for reporting the right stats and such, and some more testing to verify whether it can merge multiple groups into a single group when more than one GlobalInventoryId is associated with it.

PR #1082, opened 2025-02-12
[ChestsAnywhere] Tweak condition for big chest offset.

LeFauxMatt:
This is a small change that behaves identically to the prior logic, but fixes a future incompatibility with the Unlimited Storage Mod where chests will have a configurable width and height, and so the offset should apply to all ItemGrabMenu containing 5 rows of inventory slots.


Pathoschild (approved):

Pathoschild:
Thanks for the PR! Merged into `develop` for the upcoming Chests Anywhere 1.26.1.

PR #1088, opened 2025-03-02
Update Italian translations for Fast Animations and Lookup Anything

MakinDay:


Pathoschild:
@MakinDay To clarify this translation:
```diff
- "npc.schedule.farmhand.unknown-position": "Currently in an unknown location on the host.", // TODO
+ "npc.schedule.farmhand.unknown-position": "Attualmente in una posizione sconosciuta all'host.",
```

The intended meaning is that the location exists for the host player, but the current farmhand can't see it. In other words, it could be rephrased as "this NPC is in a location on the host, which you can't see because you're not the host".

Is the translation correct in that case? Auto-translation suggests it means it's unknown to the host instead.

Pathoschild:
Thanks for the help! Merged manually into `develop` for the upcoming Fast Animations 1.14.0 and Lookup Anything 1.50.0.

I used AI to adjust the `npc.schedule.farmhand.unknown-position` translation issue I mentioned above, but feel free to submit a new PR if it's incorrect.

PR #1092, opened 2025-03-07
Add support for Better Game Menu

KhloeLeclair:
Hello!

[Better Game Menu](https://github.com/KhloeLeclair/StardewMods/releases/tag/BetterGameMenu-Preview3) is a new mod I'm going to be releasing that replaces `GameMenu` with a replacement that:

1. Is considerably more efficient by virtue of only creating page instances when the pages are actually accessed.
2. Has a robust API to let other mods work with the game menu, making it easy to register new tabs, override existing tabs, and respond to events involving the menu.

I realize this has the potential to break a lot of things, so I'm going to be submitting PRs to various mods that interact with GameMenu to implement support for BetterGameMenu. This is such a PR.

---

Your mods don't touch on `GameMenu` much from what I can tell, with the exception of finding the current page to extract information. As such, I've stripped out almost the entire API from the `BetterGameMenuImplementation` this creates, leaving just a method to get the current page of a menu instance.

Everything was simple and easy to implement. The only note I have is that there's a logical change in Chests Everywhere to check if the current page is an `InventoryPage` instance, rather than checking which tab number is open in the `GameMenu`.

I hope this is helpful. Please contact me if you have any questions!

### DIFF
--- a/ChestsAnywhere/ModEntry.cs
+++ b/ChestsAnywhere/ModEntry.cs
@@ -99,6 +99,9 @@ internal class ModEntry : Mod
             set: config => this.Config = config
         );
 
+        // add Better Game Menu support
+        this.BetterGameMenu = new(this.Helper.ModRegistry, this.Monitor);
+
         // add Iconic Framework icon
         IconicFrameworkIntegration iconicFramework = new(this.Helper.ModRegistry, this.Monitor);
         if (iconicFramework.IsLoaded)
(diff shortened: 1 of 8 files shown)


Pathoschild (approved):

Pathoschild:
Thanks for the PR! Merged into `develop` for the upcoming Chests Anywhere 1.26.2, Debug Mode 1.15.2, and Lookup Anything 1.50.0.

PR #1100, opened 2025-03-29
[Lookup Anything] (documentation) Added an instruction for setting up multiple keyboard shortcuts

gab-umich:
Took a long time to figure this out until Chu pointed me in this direction on discord. Could be a helpful for split-screen players using gamepad + keyboard, like myself. Generic Mod Config menu doesn't allow for multi-button setup innately, so had to edit the file directly until things change. 

Putting up a quick enhancement to accept alt buttons for lookup, etc. in the config might be more elegant IMHO, but editing the file should be just fine for most folks 😄

### DIFF
--- a/LookupAnything/docs/README.md
+++ b/LookupAnything/docs/README.md
@@ -170,6 +170,17 @@ Enable tile lookups to see information about map tiles:
 ## Compatibility
 Lookup Anything is compatible with Stardew Valley 1.6+ on Linux/macOS/Windows, both single-player and
 multiplayer. There are no known issues in multiplayer (even if other players don't have it installed).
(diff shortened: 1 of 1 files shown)


Pathoschild:
Thanks for the pull request!

This should be under _Configure_ instead of _Compatibility_ though, and we should probably add an equivalent section for the other mods.

That said, I think it'd be better to add support for multi-binding options in Generic Mod Config Menu instead. I contacted the author and she's open to a pull request, so I'll look into submitting one soon.

Pathoschild:
I submitted a PR for [multi-key-binding support in Generic Mod Config Menu](https://github.com/spacechase0/StardewValleyMods/pull/502), so I'll close this one. Thanks again!

gab-umich:
Surely! Thanks for looking.

PR #1104, opened 2025-04-06
Fix off-by-one error for negative tile positions

### LINKED ISSUE #1102: TileHelper.GetTileFromCursor does not work as exepcted
Tiles that are have a negative X or Y coordinate are not returning the expected Tile coordinates.

This is reproduceable in buildings where the 0,0 coordinate is in the top left of the rendered building but extends well in the negative directions.

If you turn on DebugMode and DataLayer Grid mode you will notice the issue on any tile that is pass 0 on the X and 0 on the Y coordinates.

All negative tiles are essentially off by one because the calculation of (Game1.viewport.X + x) could return a negative value that is less than the Game1.tileSize which results in 0 instead of -1. 

Image

Here is my fix that solves the issue for me.

    public static Vector2 GetTileFromScreenPosition(float x, float y)
    {
        int tileSize = Game1.tileSize;

        int screenX = (int)(Game1.viewport.X + x);
        int screenY = (int)(Game1.viewport.Y + y);

        int tileX = screenX / tileSize;
        int tileY = screenY / tileSize;

        if (screenX % tileSize < 0 && tileX <= 0)
            tileX--;

        if (screenY % tileSize < 0 && tileY <= 0)
            tileY--;

        return new Vector2(tileX, tileY);
    }

isaacalger:
Split the calculation up a little bit so I could check if the results are negative and handle accordingly

### DIFF
--- a/Common/TileHelper.cs
+++ b/Common/TileHelper.cs
@@ -122,21 +122,12 @@ internal static class TileHelper
     /// <param name="y">The pixel Y coordinate.</param>
     public static Vector2 GetTileFromScreenPosition(float x, float y)
     {
-        int screenX = (int)(Game1.viewport.X + x);
-        int screenY = (int)(Game1.viewport.Y + y);
+        float screenX = Game1.viewport.X + x;
+        float screenY = Game1.viewport.Y + y;
 
-        int tileX = screenX / Game1.tileSize;
-        int tileY = screenY / Game1.tileSize;
-
-        if (screenX % Game1.tileSize < 0 && tileX <= 0)
-            tileX--;
-
-        if (screenY % Game1.tileSize < 0 && tileY <= 0)
-            tileY--;
+        int tileX = (int)Math.Floor(screenX / Game1.tileSize);
+        int tileY = (int)Math.Floor(screenY / Game1.tileSize);
 
         return new Vector2(tileX, tileY);
-
-        //This has issues with negative values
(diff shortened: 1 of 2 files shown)


Pathoschild:
If I understand correctly, the issue is that we truncated the decimals to get the integer tile position. That's fine for positive positions (e.g. `1.2` is over tile `1`), but it results in negative numbers incorrectly being rounded up (e.g. `-1.2` becomes `-1` even though it's actually over tile `-2`).

So instead of adjusting for that offset:
```cs
int screenX = (int)(Game1.viewport.X + x);
int screenY = (int)(Game1.viewport.Y + y);

int tileX = screenX / Game1.tileSize;
int tileY = screenY / Game1.tileSize;

if (screenX % Game1.tileSize < 0 && tileX <= 0)
    tileX--;

if (screenY % Game1.tileSize < 0 && tileY <= 0)
    tileY--;
```

I tweaked your fix to always round down instead:
```cs
float screenX = Game1.viewport.X + x;
float screenY = Game1.viewport.Y + y;

int tileX = (int)Math.Floor(screenX / Game1.tileSize);
int tileY = (int)Math.Floor(screenY / Game1.tileSize);
```

...and applied it to the position shown in Debug Mode too.

Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Data Layers 1.20.2 and Debug Mode 1.17.0.

isaacalger:
You are correct. I like your changes too it is much cleaner and easier to
understand what is going on.

Isaac Alger

PR #1106, opened 2025-04-24
[Content Patcher] patch export: Use DataLoader to guess the type of Data assets 

SinZ163:
This fixes errors when Content Patcher (or other mods) are handling the load of non-propagated vanilla assets and rely on the caller type being correct.

Example logs with this active:
https://smapi.io/log/101b626527f8496399be999bbb62bd95?Levels=trace%7Edebug%7Einfo%7Ewarn%7Eerror%7Ealert%7Ecritical&Page=8&PerPage=1000
(This was prior to the handler being case insensitive)
Initially had nothing loading Data/FishPondData so command runs fine, but I added an empty load patch to a random content pack and it stopped working when the wrong case (i.e. my logic wasn't changing the type away from object).

https://smapi.io/log/9b9ea334479443d79201663b20017311?Levels=trace%7Edebug%7Einfo%7Ewarn%7Eerror%7Ealert%7Ecritical&Page=7&PerPage=1000
(this is 1:1 what the current PR is)

### DIFF
--- a/ContentPatcher/Framework/Commands/Commands/ExportCommand.cs
+++ b/ContentPatcher/Framework/Commands/Commands/ExportCommand.cs
@@ -346,6 +346,7 @@ internal class ExportCommand : BaseCommand
     {
         IAssetName assetName = this.ContentHelper.ParseAssetName(asset);
 
+        // based on path
         if (assetName.IsDirectlyUnderPath("Maps"))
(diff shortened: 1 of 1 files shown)


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Content Patcher 2.7.0. Thanks again for the help!

PR #1107, opened 2025-04-26
[ContentPatcher] Add preview images and pages config schema

Mushymato:
Adds new features to CP config schema

Preview Images: content pack can specify a model which defines a preview image to show
- Target: asset target of a Texture2D
- SourceRect: source rectangle
- Scale: draw scale

Pages: content pack can specify a string page id
- Sections go inside pages
- The page title/description is translated with these keys:
  - `config.page.{pageId}.name`
  - `config.page.{pageId}.description`
 
Sample pack:
[[CP] Config Preview Image.zip](https://github.com/user-attachments/files/19919199/CP.Config.Preview.Image.zip)

### DIFF
--- a/ContentPatcher/Framework/GenericModConfigMenuIntegrationForContentPack.cs
+++ b/ContentPatcher/Framework/GenericModConfigMenuIntegrationForContentPack.cs
@@ -48,7 +48,7 @@ internal class GenericModConfigMenuIntegrationForContentPack : IGenericModConfig
     public void Register(TConfigMenu menu, IMonitor monitor)
     {
         // get fields by section
-        InvariantDictionary<InvariantDictionary<InvariantDictionary<ConfigField>>> fieldsByPageBySection = new() { [""] = new() };
+        InvariantDictionary<InvariantDictionary<InvariantDictionary<ConfigField>>> fieldsByPageBySection = [];
         foreach (var (name, config) in this.Config)
         {
             string pageId = config.Page?.Trim() ?? "";
@@ -195,7 +195,7 @@ internal class GenericModConfigMenuIntegrationForContentPack : IGenericModConfig
         }
 
         // image for preview image
-        if (field.PreviewImages?.Any() ?? false)
+        if (field.PreviewImages != null)
         {
             foreach (var image in field.PreviewImages)
             {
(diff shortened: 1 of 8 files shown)


ContentPatcher/Framework/GenericModConfigMenuIntegrationForContentPack.cs
```diff
@@ -33,41 +36,49 @@ internal class GenericModConfigMenuIntegrationForContentPack : IGenericModConfig
...
             section[name] = config;
         }
 
-        // add section/field elements
-        foreach ((string sectionId, InvariantDictionary<ConfigField> fields) in fieldsBySection)
-        {
-            if (!fields.Any())
-                continue;
+        // no configs to show, exit
+        if (fieldsByPageBySection.Count == 0)
```
SinZ163, line 66: collection looks like it is always at least 1 with the `""` page

ContentPatcher/Framework/GenericModConfigMenuIntegrationForContentPack.cs
```diff
@@ -182,6 +193,19 @@ private void AddField(TConfigMenu menu, string name, ConfigField field)
                 formatAllowedValue: GetValueText
             );
         }
+
+        // image for preview image
+        if (field.PreviewImages?.Any() ?? false)
```
SinZ163: This can just be a null check, as looping on an empty collection wont do anything

Pathoschild:
Thanks for the pull request! We discussed on Discord when you first submitted it, but I'll summarize here for the record.

The `ConfigSchema` is meant to define the mod's settings for the actual config logic (e.g. `config.json` and config tokens). Generating an in-game config UI using Generic Mod Config Menu (GMCM) or future mods is just a benefit of having a clear and generic config schema.

As such:
- Pages and page links are a GMCM UI concept, not part of the config schema. So I most likely won't merge that part, unless it's something that'd be useful to such a large number of content packs that it overrides the benefits of keeping to the intended purpose and scope.
- Preview images are similarly not really part of defining the config options. I'm hesitant to add it for the same reasons, but I _might_ be open to it if it's something a large number of content packs would use. I wouldn't want to add it if few content packs would use it though.

I'll mark it as draft for now in case you want to strip out the pages feature and discuss the preview images more.

I really do appreciate the PR! I suggest discussing ideas before implementing them though, just to avoid spending time developing something that may not get merged.

PR #1108, opened 2025-04-27
Preserve property references when tokens are in the name

### LINKED ISSUE #1056: [Content Patcher] EditData - Entries with tokenised keys cannot have tokenised values
**Issue**
When editing a data model, any entries with `"{{Tokens}}"` in the key will have their tokenised values left unresolved. If the value is a data model, none of its entries will be tokenised either.

Example:
```jsonc
"DynamicTokens": [
    {
        "Name": "PATH",
        "Value": "Mods/CPNestedModelTokenTest"
    },
    {
        "Name": "KEY",
        "Value": "TokenisedKey"
    }
]
```
```jsonc
// Entries:
"SimpleDictionary": {
    "UntokenisedKey_SimpleDictionary": "{{PATH}}/String",
    "{{KEY}}_SimpleDictionary": "{{PATH}}/String"
}
```
```jsonc
// Exported:
"SimpleDictionary": {
  "UntokenisedKey_SimpleDictionary": "Mods/CPNestedModelTokenTest/String",
  "TokenisedKey_SimpleDictionary": "{{PATH}}/String"
}
```

**Attachments**
Log: [log.smapi.io](https://smapi.io/log/9d7a88cb17ff40399afdf9ff9c7bc4f3)
Patch export: [json.smapi.io](https://smapi.io/json/none/589b9f4a832042b4a10d9552d5904073)
Repro solution and content pack: [CPNestedModelTokenTest.zip](https://github.com/user-attachments/files/17686060/CPNestedModelTokenTest.zip)

SinZ163:
Fixes #1056

Example patch:
```json
{
    "Action": "EditData",
    "Target": "{{PATH}}/Data",
    "Entries": {
         "SimpleDictionary": {
            "UntokenisedKey_SimpleDictionary": "{{PATH}}/String",
            "{{KEY}}_SimpleDictionary": "{{PATH}}/String"
        }
    }
}
```

When the nested key (`{{KEY}}_SimpleDictionary`) contains tokens (that are mutable) it will save a Setter method as a contextual so that when the token changes, it knows to replace the JProperty with a new one with the correct new name.

The problem is, the code for handling when the value contains a token does the same thing, though doesn't replace the JProperty it just sets the value.

When the key changes (i.e. resolves its first value) it will call that setter, replacing the JProperty and now the setter on the value is pointing to an unmounted JProperty and is for all intents and purposes a noop.

This change preserves the property reference by using reflection to set the name instead of replacing the property.

### DIFF
--- a/ContentPatcher/Framework/TokenParser.cs
+++ b/ContentPatcher/Framework/TokenParser.cs
@@ -25,8 +25,8 @@ internal class TokenParser
     /// <summary>Handles parsing raw strings into tokens.</summary>
     private readonly Lexer Lexer = Lexer.Instance;
 
-    /// <summary>Stores the Reflection accessors to the JProperty _name field</summary>
-    private static FieldInfo JPropertyName;
+    /// <summary>The private <c>_name</c> field on <see cref="JProperty"/> instances, used to update tokenized property names.</summary>
+    private static readonly FieldInfo PropertyNameField;
 
 
     /*********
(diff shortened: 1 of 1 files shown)


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Content Patcher 2.7.0. Thanks for the help!

PR #1110, opened 2025-05-02
[LookupAnything] Add internal id field

Mushymato:
image

This field is off by default. It offers cleaner view of the Id than datamining fields.

Affects every current subject that has support for "added by mod".
- Buildings (buildingType)
- Character (Name)
- Farm Animal (type)
- Item (QualifiedItemId, if flavored then QualifiedItemId/PreserveId)
- MovieSnack (Id)
- FruitTree (treeId)
- Tree (treeType)


Pathoschild:
Thanks for the PR!

As discussed on Discord, I would prefer to...

- make it always visible (since it's generally useful);
- move it below the main fields;
- and choose a format that conveys the relevant info in a player-friendly but non-ambiguous way.

For example, `(O)348/Cornucopia_BellPepperOrange` isn't an item ID; players won't be able to use it as a spawn code, paste it into config fields, etc.

I don't have a firm idea of the format yet, but it could be something like...

<table>
<tr>
<td>Item ID</td>
<td><code>348</code> (globally unique 'qualified ID': <code>(O)348</code>)</td>
</tr>
<tr>
<td>Flavor</td>
<td>🫑 <a href="#">Bell pepper</a></td>
</table>

I'll mark this as a draft for now until I can look into it more, unless you want to adjust the PR in the meantime.

Pathoschild (approved):

Pathoschild:
Thanks again! Merged into `develop` for the upcoming Lookup Anything 1.51.0.

PR #1114, opened 2025-05-09
[Content Patcher] Forward patch priorities to SMAPI instead of collapsing to Exclusive/Default

ichortower:
This PR modifies how patch priority is applied.

For Load patches, instead of selecting only the highest-priority patch on an asset and always giving it to SMAPI with `AssetLoadPriority.Exclusive`, all non-exclusive loads and up to one exclusive load are forwarded to SMAPI at the same priority level they already have. If there is more than one exclusive load for an asset, the existing behavior of erroring out now and not forwarding those patches still applies, with the same errors as before, depending on how many patches/content packs provided the loads.

For Edit patches, instead of sorting all patches by priority and giving them all to SMAPI with `AssetEditPriority.Default`, the sorting function also breaks into new groups when the current priority value changes, so each group has a consistent mod id and priority. Then each group is given to SMAPI with the priority value it has.

The effect is to allow CP patches to fully interface with the priority system, instead of working only within CP. In 2.6.1, all Content Patcher loads are Exclusive and all edits are Default, no matter what priority content pack authors specify. This prevents, e.g., a Content Patcher edit ever being able to apply after a SMAPI mod's edit that specifies a priority like `Default + 1`. With this change, if a CP pack uses `Default + 2` or `Late` or similar, it will apply after a SMAPI mod's `Default + 1`, which to me is the expected behavior. Or, for another example, if a SMAPI mod requests an `Exclusive` load on an asset, and a CP mod requests one at `Low`, in 2.6.1 SMAPI will barf because both loads are actually exclusive, which I consider a counterintuitive result. With the change, the load proceeds normally and the SMAPI load is chosen.

### DIFF
--- a/ContentPatcher/Framework/PatchManager.cs
+++ b/ContentPatcher/Framework/PatchManager.cs
@@ -461,56 +461,62 @@ internal class PatchManager
         where T : notnull
     {
         IAssetName assetName = e.NameWithoutLocale;
-        // split loaders into exclusive (must check) and non-exclusive (forward to SMAPI)
-        const int exclusivePriority = (int)AssetLoadPriority.Exclusive;
-        IPatch[] exclusiveLoaders = loaders.Where(p => p.Priority == exclusivePriority).ToArray<IPatch>();
-        IPatch[] nonExclusiveLoaders = loaders.Where(p => p.Priority != exclusivePriority).ToArray<IPatch>();
 
-        // let SMAPI handle order for non-exclusive loads, but check that their FromFiles exist
-        foreach (LoadPatch candidate in nonExclusiveLoaders)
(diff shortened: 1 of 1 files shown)


Pathoschild:
Thanks for the PR!

The new approach for `Load` patches will allocate two extra collections per load though. What's the benefit of sending multiple `Load` patches to SMAPI? Couldn't we just send the highest-priority patch like before, but set the correct priority value instead like you did for the edit patches?

Pathoschild (approved):

Pathoschild:
I adjusted it to send one loader like before, per our Discord discussion.

Merged into `develop` for the upcoming Content Patcher 1.27.0. Thanks again!

PR #1116, opened 2025-05-20
[Tractor] pickaxe: configurable clear fertilized dirt

TWT233:
### Background

For farming works, I need to reap the plants then get my seeds

Sometimes I would get my pickaxe out by mistake and the fertilized dirt got cleared immediately which is kinda frustrating

---

### This commit

added a feature that not clear fertilized dirt by default, and corresponding logic to clear them with a new config

The translations for config were generated by LLM, I can not check them all

---

### Test

manually, according to `pale / fertilized tilled dirt` & `config on / off` combinations

passed on my machine with vanilla game+mods bundle in the repo only


Pathoschild (approved):

Pathoschild:
Thanks for the pull request! Merged into `develop` for the upcoming Tractor Mod 4.23.0.

PR #1117, opened 2025-05-21
[Automate] per-storage settings

TWT233:
yeah the most difficult part is how to iter chest ids (for config showing)

there is no

        // Data/Machines
        foreach (string rawItemId in DataLoader.Machines(Game1.content).Keys)

for chests :'‑(

maybe I have to iter over all objects

### DIFF
--- a/Automate/Framework/AutomationFactory.cs
+++ b/Automate/Framework/AutomationFactory.cs
@@ -72,7 +72,9 @@ internal class AutomationFactory : IAutomationFactory
                     return new MiniShippingBinMachine(chest, location);
 
                 default:
-                    return new ChestContainer(chest, location, tile);
+                    if (chest.HasContextTag(ModConstants.StorageTag))
+                        return new ChestContainer(chest, location, tile);
+                    break;
             }
         }
 
--- a/Automate/Framework/GenericModConfigMenuIntegrationForAutomate.cs
+++ b/Automate/Framework/GenericModConfigMenuIntegrationForAutomate.cs
@@ -135,8 +135,8 @@ internal class GenericModConfigMenuIntegrationForAutomate : IGenericModConfigMen
         menu.AddDropdown(
             name: I18n.Config_DefaultChestOverride_Name,
             tooltip: I18n.Config_DefaultChestOverride_Desc,
-            get: config => config.DefaultStorageOverride.ToString(),
-            set: (config, value) => config.DefaultStorageOverride = bool.Parse(value),
+            get: config => config.ChestsEnabledByDefault.ToString(),
+            set: (config, value) => config.ChestsEnabledByDefault = bool.Parse(value),
             allowedValues: [bool.TrueString, bool.FalseString],
             formatAllowedValue: value => bool.Parse(value)
                 ? I18n.Config_ChestOverride_Values_Enabled()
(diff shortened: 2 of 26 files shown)


Pathoschild:
Thanks for the pull request!

Something like this might be good enough to get the possible chest types:
```cs
var itemRepo = new ItemRepository();
Chest[] potentialChestTypes = itemRepo
    .GetAll(ItemRegistry.type_object, includeVariants: false)
    .Concat(itemRepo.GetAll(ItemRegistry.type_bigCraftable, includeVariants: false))
    .Select(p => p.Item as Chest)
    .Where(p => p is not null)
    .ToArray()!;
```
See also `Framework/AutomationFactory.GetFor` for the logic to handle different chest types (e.g. mini-shipping bins).

Pathoschild:
I think there are two common cases players want:
- specific chests aren't automated (e.g. regular chests);
- or only specific chests are automated (e.g. the [Automators](https://www.nexusmods.com/stardewvalley/mods/15776) mod).

The PR supports the first case, but I don't think it works for the second one yet. 

We could maybe support both cases with a config UI like this:
```
Enabled chests
------------------------------------------------
First, do you want all chest types to be automated by default?
                   ┌───────────────┬─┐
Default value:     │ Automated     │▼│
                   └───────────────┴─┘

Next, you can override the above option for each chest type:
                   ┌───────────────┬─┐
Chest:             │ Default       │▼│
                   └───────────────┴─┘
                   ┌───────────────┬─┐
Big Chest:         │ Default       │▼│
                   └───────────────┴─┘
                   ┌───────────────┬─┐
Junimo Chest:      │ Not automated │▼│
                   └───────────────┴─┘
...
```
Which would get stored like this (with default values omitted):
```json
{
    "EnableChests": {
        "(BC)256": false
    }
}
```
If they set the default to 'not automated' and then enabled Junimo Chests, it would get stored like this:
```json
{
    "EnableChests": {
        "*": false,
        "(BC)256": true
    }
}
```

TWT233:
> Thanks for the pull request!
> 
> Something like this might be good enough to get the possible chest types:
> 
> ```cs
> var itemRepo = new ItemRepository();
> Chest[] potentialChestTypes = itemRepo
>     .GetAll(ItemRegistry.type_object, includeVariants: false)
>     .Concat(itemRepo.GetAll(ItemRegistry.type_bigCraftable, includeVariants: false))
>     .Select(p => p.Item as Chest)
>     .Where(p => p is not null)
>     .ToArray()!;
> ```
> 

Thanks! I'll have a try on this pattern

> See also `Framework/AutomationFactory.GetFor` for the logic to handle different chest types (e.g. mini-shipping bins).

Yes noticed `MiniShippingBinMachine` can be spawned even it's a chest

Thinking about exclude it manually, or reuse the `GetFor` with type assertion as the logic in `MachineGroupFactory.TryAddEntities`

The later would be more consistent to current logic and flexible, but would it introduce performance impact and potential problem? Since I have not checked yet how to prepare params `location` and `tile` for it, arbitraty null and any default tile are acceptable?

TWT233:
> I think there are two common cases players want:
> 
> * specific chests aren't automated (e.g. regular chests);
> * or only specific chests are automated (e.g. the [Automators](https://www.nexusmods.com/stardewvalley/mods/15776) mod).
> 
> The PR supports the first case, but I don't think it works for the second one yet.
> 
> We could maybe support both cases with a config UI like this:
> 
> ```
> Enabled chests
> ------------------------------------------------
> First, do you want all chest types to be automated by default?
>                    ┌───────────────┬─┐
> Default value:     │ Automated     │▼│
>                    └───────────────┴─┘
> 
> Next, you can override the above option for each chest type:
>                    ┌───────────────┬─┐
> Chest:             │ Default       │▼│
>                    └───────────────┴─┘
>                    ┌───────────────┬─┐
> Big Chest:         │ Default       │▼│
>                    └───────────────┴─┘
>                    ┌───────────────┬─┐
> Junimo Chest:      │ Not automated │▼│
>                    └───────────────┴─┘
> ...
> ```
> 
> Which would get stored like this (with default values omitted):
> 
> ```json
> {
>     "EnableChests": {
>         "(BC)256": false
>     }
> }
> ```
> 
> If they set the default to 'not automated' and then enabled Junimo Chests, it would get stored like this:
> 
> ```json
> {
>     "EnableChests": {
>         "*": false,
>         "(BC)256": true
>     }
> }
> ```

I'd prefer the first case for keeping consitent to current behavior of Machine?

Moreover, there are not so much types of chest(as machine), so enabled as default would be enough for user experience

BTW, maybe there could be a new PR to manage the triple state switch? It looks like not a breaking changes and I would have a look later(after this PR)

Pathoschild:
> Yes noticed `MiniShippingBinMachine` can be spawned [...] Thinking about exclude it manually, or reuse the `GetFor` with type assertion as the logic in `MachineGroupFactory.TryAddEntities`

We probably shouldn't reuse `GetFor`, since it's specialized for getting chests from the world. We could just copy the logic from `AutomationFactory.GetFor` for now; I think I'll add a chest context tag to support custom chests.

> I'd prefer the first case for keeping consitent to current behavior of Machine?

I think the second case is the more common one, but I don't mind adding that part myself.

TWT233:
> I think the second case is the more common one, but I don't mind adding that part myself.

Yes I agree

Besides is it necessary to apply this pattern for Machine too?
IMO it could be neater if we have a dedicated PR for this feature later
Otherwise I can finish it together

Pathoschild:
An option to disable all machines by default could be interesting, but I don't think it's needed for now since...
- I don't recall that feature ever being requested;
- automating machines is the main purpose of the mod;
- and players can just not connect machines to automated chests instead.

TWT233:
> An option to disable all machines by default could be interesting, but I don't think it's needed for now since...
> 
> * I don't recall that feature ever being requested;
> * automating machines is the main purpose of the mod;
> * and players can just not connect machines to automated chests instead.

ACK, that's reasonable

TWT233:
Still can not find the way to list all chest types for GMCM :'-(

Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Automate 2.4.0. Thanks for all the work!

PR #1121, opened 2025-06-03
FastAnimations: apply fishing speedup to "Hit!" and "Perfect" text animations

jltaylor-us:
Adds new handler classes for the "Hit!" and "Perfect" text animations when fishing, controlled by the same config option for fishing animation speed.


Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Fast Animations 1.15.0.

I know I suggested reusing the fishing option, but I found that in practice speeding up the text like _Hit!_ removes some of the satisfaction of fishing; so I split the option to let players choose it separately.

PR #1126, opened 2025-06-13
Fix GetTileFromScreenPosition() not acknowledging game zoom level.

AlanDavison:
Stardew's built-in Game1.currentCursorTile still breaks on negative tile co-ordinates, so fixing up the previous PR was the most sensible route.

I don't love that `Game1.options.zoomLevel` can be zero if set through code, though. I didn't want to add in checks for that and add more complexity without discussing it first. Especially since it won't hard crash, and at that point, things are already in an unplayable state.


Pathoschild (approved):

Pathoschild:
Merged into `develop` for the upcoming Debug Mode 1.17.1. Thanks for the help!

PR #1127, opened 2025-06-21
Update Chinese and Japanese translations for Tractor Mod

TWT233:
I can proofread zh / ja only so just finished these parts 🥲


Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Tractor Mod 4.23.2.

PR #1136, opened 2025-07-11
Content Patcher Profiler integration

SinZ163:
This adds support to give profiler information for individual loads and edits so that users can see the breakdown per patch rather than a single "Content Patcher took x ms" segment.

image


Pathoschild (approved):

Pathoschild:
Thanks for the PR! Merged into `develop` for the upcoming Content Patcher 2.7.2.

PR #1138, opened 2025-07-15
[Lookup Anything] Include trash bear in the "needed for" field

lisyce:
- Show users that the item being looked up needs delivered to the trash bear
- Adds one new translation key in `default.json`: `item.needed-for.trash-bear`

image

(I use BarleyZP as my modding name, so if this PR gets mentioned anywhere please use that name!)


Pathoschild (approved):

Pathoschild:
Thanks again for the help! Merged into `develop` for the upcoming Lookup Anything 1.52.0.

PR #1141, opened 2025-07-19
[Content Patcher] Update Chinese documentation

SummerFleur2997:
1. Updated multiple instances of inappropriate or overly colloquial expressions in the Chinese documentation, revising them with more formal language to match the style of technical documentation.
2. Optimized the document formatting and corrected improper punctuation usage based on Chinese punctuation standards.
3. Due to some force majeure factors, most of the top editors of the Stardew Chinese wiki have stopped updating content on the official wiki. Therefore, the original links have been replaced with those of the currently more active Chinese community wiki(hereinafter referred to as bwiki). The content in this wiki is also published under the CC-BY-NC-SA 3.0 license. As one of the currently active editors in the bwiki, we will continue to improve the content in the Modding section.

Several leading editors in our bwiki are also very keen to further improve the Chinese localization of the CP documentation and synchronize it to bwiki, bring a more convenient experience for Chinese mod developers. If you have any questions or suggestions regarding my localization work, please feel free to let us know. We would be glad to assist in further refining the localization together.

PR #1144, opened 2025-07-25
[Content Patcher] Add some comments in en documentation

SummerFleur2997:
@Mushymato and I discussed specific details regarding [pull#1141](https://github.com/Pathoschild/StardewMods/pull/1141), and proposed adding some comment in en doc.


Pathoschild:
Thanks for the PR! I think these are less useful in English though, since the code is explained in the surrounding text and directly readable for the user. It's fine for the Chinese docs to have extra code comments for clarity that aren't present in the English docs though.

So I'll close this PR for the English docs, but leave the comments in the Chinese docs.

PR #1148, opened 2025-08-14
[ContentPatcher] Add animal & building support to MigrateIds

Mushymato:
Add 2 new type options to `Pathoschild.ContentPatcher_MigrateIds`

- `FarmAnimals`: migrate farm animals owned by the player
- `Buildings`: migrate farm buildings, note that going from a building with interior map to building without map destroys the interior map.


Pathoschild (approved):

Pathoschild:
Thanks again for the help! Merged into `develop` for the upcoming Content Patcher 2.8.0.

PR #1149, opened 2025-08-18
[Tractor mod] Add option to plant tree seeds

BananaFruit476:
Late game when needing lots of wood it is tedious to plant batches of 999 tree seeds manually.

Added 2 configuration options with respective functionality:

1. Enable planting of tree seeds
2. Enable applying tree fertilizer to newly planted tree seeds if available in inventory

Both options are disabled by default

### DIFF
--- a/TractorMod/Framework/Attachments/SeedAttachment.cs
+++ b/TractorMod/Framework/Attachments/SeedAttachment.cs
@@ -1,8 +1,6 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.Xna.Framework;
-using Pathoschild.Stardew.Common.Items;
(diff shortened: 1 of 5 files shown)


Pathoschild:
Thanks for the pull request!

Omitting this feature was deliberate though. See the FAQs on the mod page:
> **Can I plant tree seeds with the tractor?**  
> Not really. You can add them as custom attachments in the mod settings, like this:
> ```json
> "CustomAttachments": [ "Acorn", "Mahogany Seed", "Maple Seed", "Pine Cone" ],
> ```
> That will plant the seeds on every valid tile around the tractor though, which isn't usually what you want. Planting seeds in a spaced pattern isn't really in scope for how the tractor works, so something like [Smart Building](https://www.nexusmods.com/stardewvalley/mods/11158) might be better for that.

This PR will plant the seeds in every available spot (instead of the spaced grid pattern they need), so many of them won't be able to grow correctly as a result.

PR #1150, opened 2025-08-28
[Content Patcher] Add step argument to Range token

Spiderbuttons:
This PR adds an optional new named argument to the Range token called "step." This argument will change how much each element in the resulting set of the range differs by, rather than the Range token always creating a list whose numbers increment strictly by 1. The step argument is optional, as mentioned, and will default to 1 if it is omitted, which means the Range token will behave as it always has for existing content packs.

For example, consider the following two uses of the Range token:
`{{Range: 1, 5}}`
`{{Range: 1, 5 |step=1}}`

Both of these will result in the following list: `1, 2, 3, 4, 5`. However, with this PR, we can now change the step argument to create a different list:
`{{Range: 1, 5 |step=2}}`

This will result in the list `1, 3, 5`, having started from the `min` input and incremented (or 'stepped') up 2 numbers for each subsequent value up to and including the `max` of 5. If we were to instead write the following:
`{{Range: 1, 5 |step=3}}`

The resulting list would be `1, 4`, as it will still respect the `max` and not exceed it.

The calculation for how many numbers are in the resulting range has of course been adjusted as well, which means the limit of 5,000 numbers in one range still applies. The following usage will work:
`{{Range: 0, 6000 |step=100}}`

This will result in a list like `0, 100, 200, 300 [...]`, which clearly won't have more than 5,000 values. However, writing `{{Range: 0, 10000 |step=2}}` will _not_ work because the resulting list would be 5,001 values long.

The documentation has also been adjusted in this PR to describe this new functionality.

### DIFF
--- a/ContentPatcher/Framework/Tokens/ValueProviders/RangeValueProvider.cs
+++ b/ContentPatcher/Framework/Tokens/ValueProviders/RangeValueProvider.cs
@@ -1,6 +1,5 @@
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
-using System.Linq;
 using ContentPatcher.Framework.Conditions;
 
 namespace ContentPatcher.Framework.Tokens.ValueProviders;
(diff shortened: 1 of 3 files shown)


Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Content Patcher 2.8.0.

PR #1153, opened 2025-10-08
[Central Station] Stardew Access Integration

khanshoaib3:
Hi,

I've just released a [beta for v1.7.0](https://github.com/stardew-access/stardew-access/releases/tag/v1.7.0-beta.1) of stardew access which will be focused on improving the support for other mods.
And I tested the beta with the central station mod, so I created this PR to ensure I'm in the right path with the changes.

To provide information to the screen reader, I've added `TileDesc` properties to certain tiles which stardew access uses ([here's a wiki on this](https://github.com/stardew-access/stardew-access/wiki/MapAccessibility)). So that is all this PR changes. Although I've not yet marked the decor tiles, like benches and tables, because ATM stardew access' object tracker cannot properly navigate through group of tiles with same names (it just considers the closest one), so until that I'll keep this PR as a draft.

Additionally, as I don't edit maps much, is it possible that I only check for the `TileDesc` property for a certain layer, say Building. Because right the mod is iterating though all the layers, for all the tiles ([look here](https://github.com/stardew-access/stardew-access/blob/f95846a96bf19f7d2d7d2f5dc0433b0c9d1768d6/stardew-access/Tiles/AccessibleTileManager.cs#L242)). Months ago when I was testing this, adding TileDesc property to different layer than the objects' just wasn't getting detected in the linked method.

Edit: `TileDesc` has been replaced with `AccessibleTilesData` data asset in v1.7.0-beta.2, will be updating the dev docs and this PR's description soon :)


khanshoaib3:
@Pathoschild Hi, what's the location/coordinates of the "blackgull" pirate? I found and described every other NPC but wasn't able to find him.

PR #1164, opened 2026-01-09
[Data Layers] Remove dependency on Automate for the machine layer

zombifier:
If Automate is not installed simply use vanilla machine detection.


Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Data Layers 1.22.0.

PR #1167, opened 2026-01-11
Update Chinese translations for Data Layers

SummerFleur2997:


Pathoschild:
* For this translation:
  ```diff
  - "fishing-depth.max": "Max depth"
  + "fishing-depth.max": "5 格以上（5）"
  ```

  The original text says "Max depth" because 5+ are all the same. For example, there's no benefit to fishing at depth 6 compared to depth 5.
* For this translation:
  ```diff
  - "fishing-depth.one": "Next to shore",
  + "fishing-depth.one": "1 格（0）",
  ```
  The original text says "Next to shore" as a hint about how depth is calculated, though that's minor.

I'll merge the translations as-is, but feel free to submit a new pull request if you want to change any to match.

Pathoschild (approved):

Pathoschild:
Thanks again for the help! Merged into `develop` for the upcoming Data Layers 1.22.0.

PR #1170, opened 2026-01-11
Update Chinese translations for Tractor Mod

SummerFleur2997:


TractorMod/i18n/zh.json
```diff
@@ -73,12 +73,11 @@
     "config.tractor-volume.name": "引擎音效大小",
     "config.tractor-volume.tooltip": "拖拉机声音效果的大小。默认值：{{defaultValue}}%。",
 
-    // TODO
-    "config.tool-use-sounds-limit.name": "Tool Use Sounds",
-    "config.tool-use-sounds-limit.tooltip": "When applying a tool to many tiles around the tractor, which tool sounds should be played. The default option is '{{defaultOption}}'. Note that '{{unlimitedOption}}' may cause errors with larger ranges.",
```
SummerFleur2997, line 78: I am not sure what does the *errors with larger ranges* refers to :(

Pathoschild: The 'larger ranges' refers to higher values for the "Distance" option, which affects how many tiles around the tractor are affected when you use a tool. If the distance is too high, it tries to play a large number of sounds at once which can cause errors.

I'll adjust the original text to be clearer:
> Note that '{{unlimitedOption}}' may cause errors if the '{{distanceOption}}' option is too high, due to playing too many sounds at once.

Pathoschild: I'll merge it as-is, but I'll add a `TODO` note after the English text is clarified.

Pathoschild (approved):

Pathoschild:
Thanks again for the help! Merged into `develop` for the upcoming Tractor Mod 4.24.3.

PR #1175, opened 2026-02-13
Fix misnamed artifact on release builds

### LINKED ISSUE #1173: [WorkFlows] A question about `.github/workflows/build.yml`
**Describe the bug**
Some issues with workflow builds that are causing ambiguity and hindering automation

**To Reproduce**
1.In [`github/workflows/build.yml`](https://github.com/Pathoschild/StardewMods/blob/develop/.github/workflows/build.yml)  line `35`
```yml
create_combined_zip: ${{github.ref == env.RELEASE_REF}}   
```
it may be true/false.
But as description of [`Pathoschild/SMAPI-ModBuildWorkflow/upload-release-artifacts@v0`](https://github.com/Pathoschild/SMAPI-ModBuildWorkflow/blob/v0/upload-release-artifacts/action.yml) line `11` 
```yml
    create_combined_zip:
        description: "The name of a combined zip file to create containing all the mod zips, or omit to disable. For example, setting this to `all mods` will upload an `all mods.zip` file. Default none."
        required: false
        default: ""

```

So I think we should fill in a meaningful name instead of true or false.
I will submit a pull request to fix this. I hope you can merge it.

---

2.The existing packaging method results in nested compression. As shown in the image below, this is the file structure after downloading and extracting a `release artifact`. This is not convenient for direct usage and distribution

Image
However, I think fixing this is quite invasive, as it requires modifying the upstream build scripts. If you're planning to tackle this issue, I'd be happy to contribute my code. Looking forward to your feedback

Lixeer:
fix #1173 ,click for details

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,7 @@
 name: Build releases
-on: push
+on: 
+  push:
+  workflow_dispatch: 
 jobs:
     build:
         runs-on: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,4 +32,4 @@ jobs:
               uses: Pathoschild/SMAPI-ModBuildWorkflow/upload-release-artifacts@v0
               with:
                   create_attestations: ${{github.ref == env.RELEASE_REF}}
-                  create_combined_zip: ${{github.ref == env.RELEASE_REF}}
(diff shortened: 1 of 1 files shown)


Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop`, so the workflow changes are now live.

PR #1177, opened 2026-02-23
Add double-tap lookup support

ferrles124:
Hi! I play Stardew Valley on Android where there's no F1 key. I added double-tap detection in ModEntry.cs so the lookup can be triggered by double-tapping on mobile.

The feature only activates on Android (GamePlatform.Android check), so it won't affect PC/other platforms.


Pathoschild (approved):

Pathoschild:
Thanks for the help! I expanded it into a feature which can be enabled on any platform (only enabled by default on Android). Merged into `develop` for the upcoming Lookup Anything 1.55.0.

PR #1187, opened 2026-03-24
Suppress navigation & deselect input on escape in Lookup Anything search UI

khanshoaib3:
- Suppresses navigation while searching with keyboard i.e., SearchTextBox is selected
- De-select search text box on Escape key press


Pathoschild (approved):

Pathoschild:
Thanks for the help! Merged into `develop` for the upcoming Lookup Anything 1.56.0.

PR #1188, opened 2026-03-25
LookupAnything: support conventional properties for mod added menus

Mushymato:
This let mods potentially avoid having to update a field value every tick since they can put that logic in the getter instead.


Pathoschild (approved):

Pathoschild:
Thanks again for the help! Merged into `develop` for the upcoming Lookup Anything 1.56.0.

